PR #72, opened 2026-01-26
Feature: Add language auto-detection and GUI selector

jatinkrmalik:
This PR implements the enhancements suggested during review of PRs #52 and #68.

### 1. Auto-detection for Whisper (`"auto"` language option)
- Adds `"auto"` as a language option for Whisper's automatic language detection
- CLI: `--language auto`
- VOSK falls back to `"en-us"` when auto is selected (VOSK doesn't support auto-detect)
- In `recognition_manager.py`: `language=None` enables Whisper's auto-detection

### 2. GUI Language Selector
- Adds dropdown in Settings dialog for language selection
- Options: Auto-detect (Whisper only), English (US), French, German, Russian
- Properly integrated with existing settings system
- Removes TODO comment about language GUI selection

### 3. Updated Documentation
- Config comments updated to mention `"auto"` option
- CLI help text updated

## Related PRs

- Builds on PR #71 (multi-language support)
- Addresses suggestions from PR #52 review
- Addresses suggestions from PR #68 review

## Testing

- [ ] Test language dropdown in Settings
- [ ] Test auto-detection with Whisper
- [ ] Test VOSK falls back to en-us with auto selected
- [ ] Test CLI `--language auto` flag

PR #68, opened 2026-01-21
Comment out language parameter in transcribe call

kappapiana:
Comment out the language parameter in the transcribe method. This will permit auto recognition of language and use the application to write languages other than hardcoded English

## Description

Using language="en" hardcodes transcribing in English even if the spoken part is in a different language

## Type of Change

- [x ] 🐛 Bug fix (non-breaking change which fixes an issue)
- [ ] ✨ New feature (non-breaking change which adds functionality)
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to change)
- [ ] 📖 Documentation update
- [ ] 🧹 Code refactoring
- [ ] ✅ Test update

## Checklist

- [ x] My code follows the code style of this project (black, isort)
- [ ] I have updated the documentation accordingly
- [ ] I have added tests to cover my changes
- [ ] All new and existing tests pass locally
- [ x] Pre-commit hooks pass

## Additional Notes

The expected behaviour is that STT transcribes the speech as it is spoken, not to default to English, this is STT + translate. It took me a lot of time to find out why it behaved so strangely.

PR #52, opened 2026-01-01
Feature: 3 more languages for Whisper and VOSK

violog:
## Description

This PR implements draft support for French (FR), German (DE), and Russian (RU) by adding configuration handling through the command line, like in the issue #13. It incorporates the code snippets (kindly proposed by @pierres52) for language selection, adds missing mappings and fallbacks, and ensures consistent behavior for these 3 languages.

Two TODOs are left:
- [ ] Support of more languages, based on [VOSK available models](https://alphacephei.com/vosk/models)
- [ ] Dynamically changing the language through GUI

I was happy to work on this part, because I really wanted working voice input on my laptop. Other contributions are welcome.

## Related Issue

Fixes #13 

## Type of Change

- [ ] 🐛 Bug fix (non-breaking change which fixes an issue)
- [x] ✨ New feature (non-breaking change which adds functionality)
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to change)
- [ ] 📖 Documentation update
- [ ] 🧹 Code refactoring
- [ ] ✅ Test update

## Checklist

- [x] My code follows the code style of this project (black, isort)
- [ ] I have updated the documentation accordingly
- [ ] I have added tests to cover my changes
- [x] All new and existing tests pass locally
- [x] Pre-commit hooks pass

PR #87, opened 2026-01-29
feat: add language selector UI with auto-detection support

jatinkrmalik:
## Summary

This PR implements a comprehensive language selector for the GUI, extending support from 4 to 10 languages and adding Whisper auto-detection.

## Changes

- **Add `SUPPORTED_LANGUAGES` constant** with 10 languages (auto, en-us, hi, es, fr, de, it, pt, ru, zh)
- **Extend `VOSK_MODEL_INFO`** to include Hindi, Spanish, Italian, Portuguese, and Chinese models
- **Add language dropdown** in Settings dialog with status indicators (✓ = Downloaded, ↓ = Will download)
- **Smart language filtering** based on selected engine:
  - VOSK: only shows languages with available models (no "auto" option)
  - Whisper: shows all supported languages including "auto" for auto-detection
- **Update default language** from `"en-us"` to `"auto"` for Whisper auto-detection
- **Update CLI `--language` argument** to include all supported languages
- **Fix install script** verification to use venv python and set `GI_TYPELIB_PATH`

## Bug Fixes

- **Fix `reconfigure()` to handle language parameter**: Previously, changing language in GUI would save to config but `self.language` was never updated, causing Whisper to continue using the old language (e.g., still transcribing in Hindi after switching to en-us)
- **Remove guard preventing language change application**: The `_processing_language_change` check in `_auto_apply_settings` prevented settings from being applied when language changed
- **Hide incompatible language options**: Only show languages that work with the selected engine, preventing user confusion

## Problem Solved

Fixes the issue where Hindi speech was being transcribed as English because:
1. Language was hardcoded to `"en-us"` in config
2. The `reconfigure()` method didn't handle the `language` parameter
3. When language is specified, Whisper transcribes in THAT language, not auto-detect

With `"auto"` language and Whisper engine:
- User speaks Hindi → Whisper auto-detects Hindi → Transcribes in Hindi ✓
- User speaks English → Whisper auto-detects English → Transcribes in English ✓

## Related PRs

- Closes #72 (supersedes draft PR with 4 languages)
- Builds on #71 (multi-language support for FR, DE, RU)
- Extends #52 (3 more languages for Whisper and VOSK)
- Addresses #68 (language parameter in transcribe call)

## Testing

- [x] Language dropdown appears in Settings dialog
- [x] Selecting "auto" works with Whisper engine
- [x] Selecting VOSK engine hides "auto" option
- [x] Switching engines auto-selects compatible language
- [x] Language changes properly apply to speech engine
- [x] Config persists language selection across restarts
- [x] CLI `--language auto` flag works
- [x] CLI `--language hi` (Hindi) flag works
- [x] All VOSK-supported languages show download status
- [x] Install script verification works correctly

### Vosk - Hindi
image

### Vosk - Russian
image

### Whisper - Spanish
image

### Whisper - German
image

> I don't speak either of last 3 languages so even with my bad accent, it was able to get it! I consider that as a win!

### DIFF
--- a/src/vocalinux/speech_recognition/recognition_manager.py
+++ b/src/vocalinux/speech_recognition/recognition_manager.py
@@ -1145,6 +1145,7 @@ class SpeechRecognitionManager:
         self,
         engine: Optional[str] = None,
         model_size: Optional[str] = None,
+        language: Optional[str] = None,
         vad_sensitivity: Optional[int] = None,
         silence_timeout: Optional[float] = None,
         audio_device_index: Optional[int] = None,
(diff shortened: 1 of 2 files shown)


jatinkrmalik:
SHIP IT!

PR #222, opened 2026-02-17
Fix #219: Suppress [BLANK_AUDIO] tokens in whisper.cpp output

### LINKED ISSUE #219: Bug: [blank audio] inserted when pausing too long with mic on
## Bug Report from Reddit User

**Source:** [Reddit comment](https://www.reddit.com/r/linux/comments/1qhogzy/comment/o5ttyu1/)
**User:** ssorbom

### Expected Behavior
When pausing with the mic on, the system should not insert any text into the text field.

### Actual Behavior
The text "[blank audio]" is being inserted into the text field every few seconds during pauses.

### Context
User has 20+ years of experience with Dragon NaturallySpeaking, so this comparison is high praise for vocalinux!

jatinkrmalik:
## Summary

This PR implements a fix for issue #219 where blank audio text was being inserted when users paused too long with the mic on.

## Root Cause

whisper.cpp outputs a special blank audio token when transcribing silent audio. Vocalinux was not filtering this out, so when a user paused with the mic on, the silence would be transcribed and inserted into the text field.

## Solution

This PR implements Option 2 from the investigation: Suppress blank audio at the whisper.cpp level by passing parameters during Model initialization.

The fix adds two parameters to the whisper.cpp Model initialization:
- suppress_blank=True - Suppress blank outputs (already the default, but set explicitly for clarity)
- suppress_non_speech_tokens=True - Suppress non-speech tokens including the blank audio token

## Changes

Modified _init_whispercpp() method in recognition_manager.py to pass the suppression parameters to the Model constructor.

## Alternative Approach

Option 1 (post-processing filter) exists as an alternative if this approach does not work as expected. That approach would involve filtering out blank audio strings after transcription completes. However, suppressing at the whisper.cpp level is cleaner and prevents the tokens from being generated in the first place.

### DIFF
--- a/src/vocalinux/speech_recognition/recognition_manager.py
+++ b/src/vocalinux/speech_recognition/recognition_manager.py
@@ -633,7 +633,12 @@ class SpeechRecognitionManager:
             cpu_count = multiprocessing.cpu_count()
 
             load_start_time = time.time()
-            self.model = Model(model_path, n_threads=n_threads)
+            self.model = Model(
+                model_path,
+                n_threads=n_threads,
+                suppress_blank=True,
+                suppress_non_speech_tokens=True,
+            )
             load_duration = time.time() - load_start_time
 
             logger.info(


jatinkrmalik:
I think it is trying to detect lack of speech and constructing something creative. Let's try this fix and test it once?

shsorbom:
`failed to initialize whisper.cpp engine: '_pywhispercpp.whisper_full_params' object has no attribute 'suppress_non_speech_tokens`

jatinkrmalik:
@shsorbom ah! I was using the older version. Fixed it in #229. Might have to make a new release now! Thank you so much for reporting back. 

Just tested on my Ubuntu 25.10 and Fedora VM: 
image

PR #251, opened 2026-02-23
fix(ibus): skip IBus setup when daemon not running

### LINKED ISSUE #250: Text injection fails on KDE Plasma / non-GNOME desktops (IBus daemon not running)
## Summary

On desktop environments like KDE Plasma (Fedora, CachyOS, etc.), the IBus daemon is not started by default. This causes Vocalinux's IBus-based text injection to fail silently or with errors, even though speech recognition works correctly.

This issue was originally reported mixed with audio issues in #34, but it's a separate problem worth tracking independently.

## Affected Environments

- Fedora KDE Plasma (Wayland)
- CachyOS (KDE)
- Other non-GNOME desktops where IBus is not the default input method

## Root Cause

1. **GNOME** automatically starts `ibus-daemon` for all users via the session manager
2. **KDE Plasma** uses `imsettings` framework, which explicitly ignores IBus on KDE Wayland (`IMSETTINGS_IGNORE_SESSION=KDE-wayland` in `/etc/X11/xinit/xinput.d/ibus.conf`)
3. When Vocalinux tries to register its IBus engine with no daemon running, it fails

## Symptoms

Users see errors like:
```
ERROR | Failed to register IBus engine
```

Or:
```
IBUS-CRITICAL: ibus_bus_get_global_engine: assertion 'IBUS_IS_BUS (bus)' failed
No engine is set.
```

## Solution

**PR in progress**: We're adding detection for whether `ibus-daemon` is running before attempting IBus setup. If not running, Vocalinux will:
1. Skip IBus entirely (no failed registration attempts)
2. Log an informative message explaining the situation
3. Fall back to ydotool/xdotool automatically

## Workaround (for users who want IBus)

From @treedy in #34:
```bash
export IBUS_COMPONENT_PATH=${IBUS_COMPONENT_PATH}:~/.local/share/ibus/component/
ibus-daemon -drx
```

Then restart Vocalinux.

## Related

- #34 (original mixed thread)
- #240 (fix for IBus setup exception not triggering fallback)
- #164 (similar IBus issues)

AntoineMontane:
## Summary

On desktop environments like KDE Plasma (Fedora, CachyOS, etc.), the IBus daemon is not started by default. Previously, Vocalinux would attempt to register its IBus engine anyway, leading to failures and confusing error messages.

## Changes

- Added `is_ibus_daemon_running()` function to check if `ibus-daemon` is running
- Modified `_check_dependencies()` to skip IBus setup when daemon isn't running
- Added informative log message explaining why IBus is skipped and linking to documentation

## Before

```
ERROR | Failed to register IBus engine
INFO | Using IBus for wayland-ibus text injection
# Then text injection fails silently
```

## After

```
INFO | IBus daemon not running. This is normal on some desktop environments (e.g., KDE Plasma). Using alternative text injection method. For IBus setup, see: https://github.com/jatinkrmalik/vocalinux/wiki/IBus-Setup
INFO | Using ydotool for Wayland text injection
# ydotool fallback works immediately
```

## Test plan

- [x] Test on Ubuntu/GNOME (IBus daemon running) - should use IBus as before
- [ ] Test on Fedora KDE (no IBus daemon) - should skip IBus and use ydotool
- [ ] Verify log message appears when IBus is skipped

Fixes #250

### DIFF
--- a/src/vocalinux/text_injection/ibus_engine.py
+++ b/src/vocalinux/text_injection/ibus_engine.py
@@ -111,6 +111,28 @@ def is_ibus_available() -> bool:
     return IBUS_AVAILABLE
 
 
+def is_ibus_daemon_running() -> bool:
+    """
+    Check if the IBus daemon (ibus-daemon) is currently running.
+
+    This is important because on some desktop environments (e.g., Fedora KDE),
+    the IBus daemon is not started by default. Attempting to set up IBus
+    when the daemon isn't running will fail.
+
+    Returns:
+        True if ibus-daemon is running, False otherwise
+    """
+    try:
+        result = subprocess.run(
+            ["pgrep", "-x", "ibus-daemon"],
+            capture_output=True,
+            timeout=2,
+        )
+        return result.returncode == 0
+    except (subprocess.SubprocessError, FileNotFoundError):
+        return False
+
+
 def _get_expected_component_xml() -> str:
     """Generate the expected component XML content for the current installation."""
     engine_script = Path(__file__).resolve()
--- a/src/vocalinux/text_injection/text_injector.py
+++ b/src/vocalinux/text_injection/text_injector.py
@@ -16,6 +16,7 @@ from typing import Optional  # noqa: F401
 from .ibus_engine import (
     IBusTextInjector,
     is_ibus_available,
+    is_ibus_daemon_running,
 )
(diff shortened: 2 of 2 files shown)


jatinkrmalik:
LGTM! Thank you.

PR #271, opened 2026-02-27
Remove copy, cut, and paste from voice action commands

christiantwu:
## Summary
- Removes "copy", "cut", and "paste" from `action_commands` in `CommandProcessor`, so these common English words are no longer intercepted as keyboard shortcut actions (`Ctrl+C`/`Ctrl+X`/`Ctrl+V`)
- Previously, any dictated speech containing these words would trigger the shortcut and silently discard text spoken before the command word
- Removes associated hardcoded test cases that referenced these commands

## Test plan
- [x] Verified that "copy", "cut", and "paste" now pass through as regular dictated text
- [x] Verified that existing action commands ("delete that", "undo", etc.) continue working
- [x] Tested with full sentences containing these words (e.g., "I need a copy of that document")

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -124,6 +124,9 @@ jspm_packages/
 # Temporary folders
 tmp/
 temp/
+MagicMock/
+tmp/
+temp/
 
 # Editor directories and files
 .vscode/*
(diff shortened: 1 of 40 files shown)


christiantwu:
@jatinkrmalik Thanks for catching that!

Issue #292, opened 2026-03-06
[BUG] Keyboard layout set to English after launching vocalinux-gui


jgbarah:
## 🐛 Bug Description

I have a Spanish keyboard. When I launch vocalinux-gui, it seems to switch the keyboard layout to English: keys produce characters as in the English keyboard, in all applications. Besides, if I set vocalinux-gui to lauch at startup, the keyboard layout seems to be Spanish, except that dead keys don' t work, Curiously enough, it is fixed is I run ibus restart (but this doesn' t fix the English layout thing if I didn' t launch vocalinux-gui at startup)..

## 📋 Steps to Reproduce

1. Boot as usual, with Spanish layout.
2. Launch vocalinux-gui
3. In any application keyboard layout seems to be English now.

## ✅ Expected Behavior

The keyboard layout to remain in Spanish.

## ❌ Actual Behavior

The keyboard layout seems to switch to English.

## 💻 Environment
- **OS**: Debian testing
- **Python Version**: 3.13
- **Vocalinux Version**: 0.8.0/beta
- **Display Server**: X11
- **Speech Engine**: Whisper_CPP on Vulkan
- **Model Size**: Tiny

## 📋 Debug Logs
Run `vocalinux --debug` and paste relevant log output:
```
$ vocalinux --debug
2026-03-06 21:21:42,352 - vocalinux.single_instance - INFO - Single instance lock acquired
2026-03-06 21:21:42,354 - vocalinux.main - DEBUG - Debug logging enabled
2026-03-06 21:21:42,471 - vocalinux.utils.resource_manager - DEBUG - Checking resources candidate: /home/jgb/.local/share/vocalinux/venv/lib/python3.13/resources (exists: False)
2026-03-06 21:21:42,471 - vocalinux.utils.resource_manager - DEBUG - Checking resources candidate: /home/jgb/.local/share/vocalinux/venv/share/vocalinux/resources (exists: True)
2026-03-06 21:21:42,471 - vocalinux.utils.resource_manager - DEBUG - Checking resources candidate: /home/jgb/.local/share/vocalinux/resources (exists: False)
2026-03-06 21:21:42,471 - vocalinux.utils.resource_manager - DEBUG - Checking resources candidate: /home/jgb/.nix-profile/bin/vocalinux/resources (exists: False)
[...]

Issue #199, opened 2026-02-12
Keyboard layout mismatch: English speech recognized but typed as French AZERTY


AntoineMontane:
## Problem

When dictating in English, the text appears as if typed on a French AZERTY keyboard layout instead of QWERTY.

**Example:**
- Speaking: "I'm talking in English but the later appeared as if it were written on a French keyboard"
- Output: "Iù, tqlking in English; but the lqter qppeqred qs if it zere zritten on q French keyboqrd"

**Character mappings observed:**
- `a` → `q`
- `q` → `a`
- `w` → `z`
- `z` → `w`
- `'` → `ù` (apostrophe becomes ù)
- `;` → `,` (semicolon issues)

## Expected behavior

Text should be output using the correct keyboard layout matching the system's input method, regardless of physical keyboard layout.

## Environment

- OS: Linux
- Keyboard: Physical AZERTY but expecting QWERTY output (or vice versa)

## Possible causes

1. Vocalinux is sending keycodes instead of characters
2. XKB layout not being respected
3. Mismatch between speech recognition language and keyboard simulation method

AntoineMontane:
Hey; no proble, to shqre ,y log zith you: ;-)

```
VocaLinux Logs - Copied at 2026-02-13 02:55:45
================================================================================

[02:53:13.972] INFO     vocalinux.ui.logging_manager | Logging manager initialized
[02:53:13.973] INFO     vocalinux.ui.logging_manager | Global logging manager initialized
[02:53:13.973] INFO     vocalinux.main       | Logging system initialized
[02:53:13.973] INFO     vocalinux.ui.config_manager | Loaded configuration from /home/antoine/.config/vocalinux/config.json
[02:53:13.973] INFO     vocalinux.main       | Using engine=whisper_cpp (from saved config)
[02:53:13.973] INFO     vocalinux.main       | Using language=en-us (from saved config)
[02:53:13.973] INFO     vocalinux.main       | Using model=base (from saved config)
[02:53:13.973] INFO     vocalinux.main       | Final settings: engine=whisper_cpp, language=en-us, model=base
[02:53:13.974] INFO     vocalinux.main       | Initializing Vocalinux...
[02:53:13.975] INFO     vocalinux.speech_recognition.recognition_manager | Initializing speech recognition with whisper_cpp engine, en-us language and base model
[...]

Issue #164, opened 2026-02-08
[BUG]  ydotool types wrong characters on non us keyboard layouts


christophtschmidt:
## 🐛 Bug Description
I'm using german keyboard layout. ydotool is typing some keys like y, z, ä, ö etc. wrong.
I know this is an ydotool issue. But theres no other option on KDE Plasma with Wayland..

## 📋 Steps to Reproduce
Simly speak a text containing characters mentioned above. 

## ✅ Expected Behavior
characters are typed the way they are recognized.
maybe implement vocalinux as  input method (IME)?

## 💻 Environment
- **OS**: Fedora 42
- **Python Version** 3.13.11
- **Vocalinux Version**: 0.15.0
- **Display Server**:  Wayland
- **Speech Engine**:tested [VOSK and  Whisper
- **Model Size**: tested small and medium 

```

## 📝 Additional Context
In vocalinux log the recognized text is shown with  the correct characters.

christophtschmidt:
and btw: i needed to fix permission issues on Fedora to get ydotool working
https://discussion.fedoraproject.org/t/ydotool-not-working-on-fedora-40/114333/4

mr-bembel:
It works great in the Vocalinux test box now. But still EN input  in various apps...

AntoineMontane:
@christophtschmidt .local/share/ibus/component/ is not used on gnome either by default ; the ibus daemon is restarted with a modified PATH https://github.com/jatinkrmalik/vocalinux/blob/f79917c386a7be9312e65bff9ffcec81e188e832/src/vocalinux/text_injection/ibus_engine.py#L720C14-L720C33
Could you provide your logs ?

christophtschmidt:
Hi :-)

The output of the commands is in the txt:
[shell_output.txt](https://github.com/user-attachments/files/25458114/shell_output.txt)

AntoineMontane:
Great !

christophtschmidt:
I need to run "ibus restart" every time i restart vocalinux - else there's a second ibus process hanging that blocks everything.

fsioni:
Heads-up that the non-US garbling isn't fully resolved: #362 added clipboard paste only for **non-ASCII** text, but **ASCII** still goes through `ydotool type`, which emits positional (US-QWERTY) keycodes and so scrambles plain ASCII on AZERTY/QWERTZ/Dvorak (e.g. `message` → `,essqge`, `avec` → `qvec`).

[...]

PR #304, opened 2026-03-14
fix(text-injection): use register_component for IBus engine activation

### LINKED ISSUE #303: [BUG] IBus engine activation times out on standalone launch — falls back to xdotool, garbling non-US   layouts - Fix in PR #304
Human disclaimer, the issue below has been generated in collaboration with Claude Opus 4.6. I have personally worked with it to debug, diagnose and fix the issue, while challenging the perceptions and looking for potential flaws in reasoning/hallucinations. With that said, I personally have little to no experience with IBus before today. I believe the associated fix resolves the issue and does not introduce any unintended consequences. 

## Bug Description

The IBus text injection path never successfully activates. The engine process starts and connects to the IBus D-Bus, but uses `bus.request_name()` to register — a method that only works when IBus itself launched the process. Since Vocalinux launches the engine directly via `start_engine_process()`, it must instead call `bus.register_component()` to announce the engine factory to IBus. Without this, `set_global_engine('vocalinux')` times out because IBus has no factory to route to, and Vocalinux falls back to xdotool.

On non-US keyboard layouts (Dvorak, AZERTY, etc.), the xdotool fallback sends QWERTY keycodes that get remapped through the active layout, producing garbled text. On Wayland, xdotool operates through XWayland with additional keymap mismatches.

A secondary issue: the `<exec>` line in the component XML and `main()` do not handle the `--ibus` and `--xml` flags that are part of the standard IBus engine protocol (as used by ibus-typing-booster and other engines).

**Related:** The IBus engine was introduced in PR #211 by @AntoineMontane — excellent work that solved the fundamental problem of layout-independent text injection. The `register_component` issue is a refinement: the engine registration works on some system configurations but not others, depending on whether IBus can discover the standalone-launched factory.

See also: #164 (ydotool wrong characters on non-US layouts), #199 (AZERTY keyboard mismatch), #292 (keyboard layout switched to English on launch). These all stem from the IBus engine not activating properly on certain configurations, forcing fallback to keycode-based injection which breaks on non-US layouts.

**A fix has been found and a PR is being prepared.**

## Steps to Reproduce

1. Install Vocalinux on a system with IBus (e.g., Fedora with GNOME/Wayland)
2. Set a non-US keyboard layout (e.g., US Dvorak)
3. Start Vocalinux and activate voice dictation
4. Observe garbled text output (e.g., "hello" becomes "d.llr" on Dvorak)

## Expected Behavior

Vocalinux should register its IBus engine, activate it, and inject text directly via IBus `commit_text()`, bypassing keyboard layout remapping.

## Actual Behavior

- Engine activation times out (`request_name` instead of `register_component`)
- Falls back to xdotool, which produces garbled text on non-US layouts
- On Wayland, xdotool fallback has additional issues due to XWayland keymap mismatches

## Environment

- **OS**: Fedora 43 (GNOME/Wayland)
- **Python Version**: 3.14
- **Vocalinux Version**: 0.8.0-beta
- **Display Server**: Wayland
- **Speech Engine**: whisper_cpp
- **Keyboard Layout**: US Dvorak

### Engine activation timeout
```
vocalinux.text_injection.ibus_engine - INFO - Starting IBus engine process: .../ibus_engine.py
vocalinux.text_injection.ibus_engine - INFO - IBus engine process started successfully
vocalinux.text_injection.ibus_engine - INFO - Activating Vocalinux IBus engine...
vocalinux.text_injection.ibus_engine - ERROR - Failed to switch IBus engine: Command '['ibus', 'engine', 'vocalinux']' timed out after 5 seconds
vocalinux.text_injection.text_injector - WARNING - IBus initialization failed: Failed to activate Vocalinux IBus engine.
```

### xdotool fallback with garbled output
```
vocalinux.text_injection.text_injector - INFO - Text injected using xdotool: ' The quick brown fox...' (45 chars)
```
Actual screen output: `Yd. 'gcct xpr,n frq hgmp.e rk.p yd. RK: erg.`

## Root Cause

The IBus protocol has two registration paths, following the same pattern as ibus-typing-booster and other Python IBus engines:

- **Launched by IBus** (`--ibus` flag): call `bus.request_name()` — IBus already knows about the component from its XML cache
- **Launched standalone** (by the application): call `bus.register_component()` — the engine must announce itself to IBus with a full component and engine description

Vocalinux always launches the engine itself via `start_engine_process()` but uses the IBus-launched registration path (`request_name` only). The engine process connects to IBus and creates a factory, but IBus doesn't know about it. When `set_global_engine('vocalinux')` is called, IBus has no factory to route to and the call times out after 5 seconds.

Additionally, the component XML `<exec>` line does not include `--ibus`, and `main()` does not handle the `--xml` or `--ibus` flags that IBus passes to engine executables. While the current inline component XML means IBus reads engine definitions directly from disk (so `--xml` is not strictly required for discovery), handling these flags correctly follows the established IBus engine protocol and enables future compatibility with exec-based discovery.

## Additional Context

- The `register_component` vs `request_name` pattern is the established approach used by ibus-typing-booster (maintained by Red Hat, shipped with Fedora) and other Python IBus engines. The API has been stable since IBus 1.4 (2012); all current distributions ship IBus 1.5.26+.
- The xdotool fallback works correctly on US QWERTY layouts, which likely explains why this issue went undetected — it only manifests with non-US keyboard layouts or on Wayland where xdotool has additional limitations.
- On Wayland without IBus, the fallback chain is: ydotool (permission denied without socket access) → wtype (compositor doesn't support virtual keyboard protocol) → xdotool via XWayland (works but garbles non-US layouts).

GrahamJenkins:
Human disclaimer: This PR implements a fix that I found with the assistance of Claude Opus 4.6. As noted in issue #303 I personally have zero prior experience with IBus and am relying on AI debugging and diagnosis. I believe this is a valid fix, and I have taken care to try to make sure this has no cross-distro side effects. If there are any mistakes or hallucinations with this PR feel free to ping me and I will chastise the agent accordingly. This fix is in place on my personal system (Fedora 43 Workstation) and allows Vocalinux to work properly for me with no known system modification required.

## Description

Fix IBus engine activation so that text injection works correctly on non-US keyboard layouts and Wayland.

The core issue: when Vocalinux launches the IBus engine process itself (via `start_engine_process()`), the engine called `bus.request_name()` to register with IBus. This is only sufficient when IBus launched the process and already knows about the engine. Standalone-launched engines must call `bus.register_component()` with a full component/engine description so IBus can discover the factory. Without this, `set_global_engine('vocalinux')` timed out because IBus had no factory to route to, causing a fallback to xdotool which garbles text on non-US layouts.

The fix follows the same pattern used by ibus-typing-booster (the reference Python IBus engine shipped with Fedora):
- **`--ibus` flag**: when IBus launches the engine, use `request_name()` (it already knows the component)
- **No flag** (Vocalinux launches it): use `register_component()` to announce the engine
- **`--xml` flag**: print the `<engines>` XML block to stdout and exit (standard IBus engine discovery protocol)
- **`<exec>` line**: updated to include `--ibus` so IBus-launched processes take the correct path

Connection guards (`is_connected()` / `get_connection()` checks) were also added so the engine process fails fast with a clear error instead of silently registering a broken factory.

## Related Issue

Fixes #303

This is a refinement of the IBus engine introduced in PR #211 by @AntoineMontane. The original implementation works on some system configurations, but on others (e.g., Fedora 43 / GNOME Wayland) the standalone-launched engine process isn't discoverable by IBus without `register_component`.

See also: #164, #199, #292 — related issues where non-US layouts produce garbled text due to the IBus engine not activating on certain configurations.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

## Changes

- Change `VocalinuxEngineApplication.__init__` to accept `exec_by_ibus` parameter:
  - `True` (IBus launched): `bus.request_name()` (existing behavior)
  - `False` (Vocalinux launched): `bus.register_component()` with full component/engine description
- Add `--ibus` argv handling in `main()`, passed through to `VocalinuxEngineApplication`
- Add `--ibus` to the `<exec>` line in `_get_expected_component_xml()` so IBus-launched processes use `request_name`
- Add `_get_engines_xml()` and `--xml` argv handling in `main()` for IBus engine discovery protocol compliance
- Add `is_connected()` and `get_connection()` guards with clear error messages
- Add `register_component()` return value check

## Testing

- Verified on Fedora 43, GNOME/Wayland, US Dvorak layout
- Engine activation succeeds without timeout (previously timed out after 5s)
- End-to-end: dictated text appears correctly on Dvorak (previously garbled via xdotool fallback)
- `ibus_engine.py --xml` outputs valid `<engines>` XML and exits immediately
- `ibus write-cache && ibus list-engine | grep vocalinux` finds the engine
- Existing unit tests pass

## Checklist

- [ ] I have updated the documentation accordingly (no user facing documentation required)
- [x] Pre-commit hooks pass

## Additional Notes

- The `register_component()` / `request_name()` pattern is identical to what ibus-typing-booster uses (maintained by Red Hat). The API has been stable since IBus 1.4; all current distributions ship IBus 1.5.26+.
- These changes are backwards-compatible: `--xml` and `--ibus` handlers only activate when those flags are explicitly passed, and `exec_by_ibus` defaults to `False`.
- The `--xml` handler is a protocol compliance measure. With the current inline component XML, IBus reads engine definitions directly from disk. However, supporting `--xml` follows the standard IBus engine convention and enables exec-based discovery if needed in the future.

### DIFF
--- a/src/vocalinux/text_injection/ibus_engine.py
+++ b/src/vocalinux/text_injection/ibus_engine.py
@@ -68,6 +68,24 @@ ENGINE_LONGNAME = "Vocalinux"
 ENGINE_DESCRIPTION = "Vocalinux voice dictation (use as default input method)"
 COMPONENT_NAME = "org.freedesktop.IBus.Vocalinux"
 
+# Shared metadata used by component XML, --xml output, and runtime registration.
+# Kept in one place to prevent drift between the three consumers.
+ENGINE_RANK = 50
+_ENGINE_META = {
+    "language": "other",
+    "license": "GPL-3.0",
+    "author": "Vocalinux",
+    "icon": "audio-input-microphone",
+    "layout": "default",
+}
+_COMPONENT_META = {
+    "version": "1.0",
+    "license": "GPL-3.0",
+    "author": "Vocalinux",
+    "homepage": "https://github.com/jatinkrmalik/vocalinux",
+    "textdomain": "vocalinux",
+}
+
 
 def ensure_ibus_dir() -> None:
     """Ensure the IBus data directory exists with secure permissions."""
@@ -210,32 +228,38 @@ def start_ibus_daemon():
         return False
 
 
+def _get_exec_command() -> str:
+    """Return the engine exec command for component XML and runtime registration."""
+    engine_script = Path(__file__).resolve()
+    return f"{sys.executable} {engine_script} --ibus"
+
+
 def _get_expected_component_xml() -> str:
     """Generate the expected component XML content for the current installation."""
-    engine_script = Path(__file__).resolve()
-    python_exec = sys.executable
+    e = _ENGINE_META
+    c = _COMPONENT_META
 
     return f"""<?xml version="1.0" encoding="utf-8"?>
 <component>
     <name>{COMPONENT_NAME}</name>
     <description>{ENGINE_DESCRIPTION}</description>
-    <exec>{python_exec} {engine_script} --ibus</exec>
-    <version>1.0</version>
-    <author>Vocalinux</author>
-    <license>GPL-3.0</license>
-    <homepage>https://github.com/jatinkrmalik/vocalinux</homepage>
-    <textdomain>vocalinux</textdomain>
+    <exec>{_get_exec_command()}</exec>
+    <version>{c['version']}</version>
+    <author>{c['author']}</author>
+    <license>{c['license']}</license>
+    <homepage>{c['homepage']}</homepage>
+    <textdomain>{c['textdomain']}</textdomain>
     <engines>
         <engine>
             <name>{ENGINE_NAME}</name>
-            <language>other</language>
-            <license>GPL-3.0</license>
-            <author>Vocalinux</author>
-            <icon>audio-input-microphone</icon>
-            <layout>default</layout>
             <longname>{ENGINE_LONGNAME}</longname>
+            <language>{e['language']}</language>
+            <license>{e['license']}</license>
+            <author>{e['author']}</author>
+            <icon>{e['icon']}</icon>
+            <layout>{e['layout']}</layout>
             <description>{ENGINE_DESCRIPTION}</description>
-            <rank>50</rank>
+            <rank>{ENGINE_RANK}</rank>
         </engine>
     </engines>
 </component>
@@ -644,28 +668,24 @@ class VocalinuxEngineApplication:
         if exec_by_ibus:
             # IBus launched us — it already knows about our component,
             # just claim the well-known D-Bus name.
-            self.bus.request_name(COMPONENT_NAME, 0)
+            if not self.bus.request_name(COMPONENT_NAME, 0):
+                logger.error("bus.request_name() failed")
+                raise RuntimeError("Failed to acquire IBus D-Bus name")
         else:
             # Launched by Vocalinux directly — register the full component
             # so IBus discovers our engine without having launched us.
             component = IBus.Component(
                 name=COMPONENT_NAME,
                 description=ENGINE_DESCRIPTION,
-                version="1.0",
-                license="GPL-3.0",
-                author="Vocalinux",
-                homepage="https://github.com/jatinkrmalik/vocalinux",
-                textdomain="vocalinux",
+                command_line=_get_exec_command(),
+                **_COMPONENT_META,
             )
             engine_desc = IBus.EngineDesc(
                 name=ENGINE_NAME,
                 longname=ENGINE_LONGNAME,
                 description=ENGINE_DESCRIPTION,
-                language="other",
-                license="GPL-3.0",
-                author="Vocalinux",
-                icon="audio-input-microphone",
-                layout="default",
+                rank=ENGINE_RANK,
+                **_ENGINE_META,
             )
             component.add_engine(engine_desc)
             if not self.bus.register_component(component):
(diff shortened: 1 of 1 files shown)


jatinkrmalik:
Thanks for this fix @GrahamJenkins! The approach is solid and aligns with how ibus-typing-booster handles the same problem.

**What I like:**
- Clean separation of `--ibus` vs standalone launch paths
- Connection guards with clear error messages
- `--xml` flag for IBus discovery protocol compliance
- Backwards-compatible (old installs without `--ibus` fall back gracefully)

**A few non-blocking suggestions for a follow-up:**

1. **Runtime Component Metadata** (line 651): The `IBus.Component` built at runtime is missing `command_line`/`exec`. The installed XML has it, so current behavior works, but it would be cleaner to keep runtime and cache metadata identical. Consider either loading from the same XML or setting `command_line` explicitly.

2. **Check `request_name()` return** (line 647): Currently the return value is ignored. If name acquisition fails silently, debugging in the field gets harder.

3. **Metadata consolidation**: Engine metadata now exists in three places (`_get_expected_component_xml()`, `_get_engines_xml()`, and the runtime `IBus.Component`). A shared builder would prevent future drift.

None of these should block merge, this is a good fix for the Wayland/non-US layout issues.  Let me know if you want to address them as well in the same pull request?

PR #321, opened 2026-03-15
fix: implement real-time streaming transcription

### LINKED ISSUE #320: [BUG] Real Time Transcription doesn't work.
Environment: Debian Trixie/KDE Wayland

`vocalinux --debug` output: https://pastebin.com/uXLZzK2K

On a side note, while the app still has some quirks, overall it seems to work fine. One another voice-to-text project that seems to care about Linux is Openwhispr. But it doesn't support real-time transcription. So, if this is fixed, it would be great differentiator.

Thanks for open-sourcing this project.

saschabuehrle:
## Bug
Fixes #320 — Real Time Transcription doesn't work.

The issue was that the current implementation waits for silence periods before processing audio, which is not truly "real-time". Users expect continuous transcription as they speak, not batch processing after pauses.

## Fix
This PR implements true real-time streaming transcription by:

- **Added real_time_streaming config option** (default: true) - enables streaming mode
- **Added streaming_chunk_duration config option** (default: 3.0s) - controls chunk size  
- **Modified recognition loop** to process audio chunks at regular intervals instead of only on silence
- **Preserved existing behavior** when streaming is disabled for backward compatibility

## How it works
- **Before**: Audio → Wait for silence → Process entire buffer → Transcribe
- **After**: Audio → Process every 3 seconds → Transcribe continuously

The system now processes audio chunks every 3 seconds while the user is speaking, providing true real-time feedback instead of waiting for silence periods.

## Testing
Validated the streaming logic with unit tests to ensure proper chunk timing and processing intervals.

Happy to address any feedback.

Greetings, saschabuehrle

### DIFF
--- a/src/vocalinux/main.py
+++ b/src/vocalinux/main.py
@@ -303,6 +303,8 @@ def main():
 
     vad_sensitivity = saved_settings.get("vad_sensitivity", 3)
     silence_timeout = saved_settings.get("silence_timeout", 2.0)
+    real_time_streaming = saved_settings.get("real_time_streaming", True)
+    streaming_chunk_duration = saved_settings.get("streaming_chunk_duration", 3.0)
     voice_commands_enabled = saved_settings.get("voice_commands_enabled")  # None = auto
     audio_device_index = audio_settings.get("device_index", None)
 
@@ -321,6 +323,8 @@ def main():
             language=language,
             vad_sensitivity=vad_sensitivity,
             silence_timeout=silence_timeout,
+            real_time_streaming=real_time_streaming,
+            streaming_chunk_duration=streaming_chunk_duration,
             voice_commands_enabled=voice_commands_enabled,
             audio_device_index=audio_device_index,
         )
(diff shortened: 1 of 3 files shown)


amaramrahul:
Thanks for looking into this. Quick question / feedback. 3sec might not really feel real-time. Is there any reason for why not to use a smaller duration such as 500ms?

Also, maybe a dumb question, but I am assuming that I can continue talking as the transcription happens real-time and there would no loss of my words because the process is busy transcribing.

saschabuehrle:
Good point! I chose 3 seconds as a balance between responsiveness and processing efficiency, but you're absolutely right that it should be configurable.

Looking at the code, the duration is already configurable via the `streaming_chunk_duration` setting in the config file (~/.config/vocalinux/config.json). You can set it to 0.5 seconds like this:

```json
{
  "speech_recognition": {
    "streaming_chunk_duration": 0.5
  }
}
```

And yes, you can continue talking while transcription happens - the audio stream is processed in chunks but the recording is continuous, so no words are lost.

Would you like me to add a UI setting to make this more discoverable, or does the config file approach work for you?

Greetings, saschabuehrle

jatinkrmalik:
@saschabuehrle I would expect you to at least reply with your own brain and thinking. It's fine to use LLM models to write code as long as you understand what they are doing. This is a big red flag.

jatinkrmalik:
> Understood.
> 
> I’ll stop pushing changes here unless there’s specific feedback on the implementation itself. If you’d rather not proceed with this PR, feel free to close it.
> 
> Greetings, saschabuehrle

@saschabuehrle I am not trying to be rude, I just want you to take ownership of your pull request. That's all.

I like the change, it seems like but the pipeline is failing with some error. If you want to fix it, you can update it and I'm happy to review it.

jatinkrmalik:
Continuing the work here on #387 since author decided to close the PR. Using a bit diff approach as well.

PR #339, opened 2026-03-23
fix(evdev): handle SYN_DROPPED to prevent stale modifier state

### LINKED ISSUE #338: [BUG] evdev backend does not handle SYN_DROPPED events
## 🐛 Bug Description

The evdev keyboard backend (`evdev_backend.py`) does not handle `SYN_DROPPED` events from the kernel. When the kernel's evdev client buffer overflows, it sends `SYN_DROPPED` to signal that events were lost. Per the [kernel input documentation](https://www.kernel.org/doc/html/latest/input/event-codes.html):

> "Client should ignore all events up to and including next SYN_REPORT event and query the device to obtain its current state."

Without handling this, a lost key-release event leaves the modifier key in a "stuck" pressed state in `key_pressed_devices`. For a push-to-talk daemon this means recording may never stop (or never start).

## 📋 Steps to Reproduce

1. Run Vocalinux with evdev backend
2. Cause high system load (e.g. heavy whisper.cpp transcription + other I/O)
3. The evdev monitor thread stalls long enough for the kernel buffer to overflow
4. A `SYN_DROPPED` is sent — the backend ignores it and processes stale/corrupt events

## ✅ Expected Behavior

Events between `SYN_DROPPED` and the next `SYN_REPORT` should be discarded, and modifier key state (`key_pressed_devices`) should be reset to prevent stuck keys.

## ❌ Actual Behavior

All events are processed regardless of `SYN_DROPPED`, potentially causing stale modifier state.

## 📝 Additional Context

- The [libevdev SYN_DROPPED guide](https://www.freedesktop.org/software/libevdev/doc/latest/syn_dropped.html) describes the required handling in detail
- `python-evdev` exposes `SYN_DROPPED` as a constant but does not handle it automatically — applications must implement their own logic
- This is an edge case (keyboard events are sparse, buffer is 64+ events), but the fix is trivial (~15 lines) and prevents a catastrophic failure mode for push-to-talk users

Sbenazar:
## Description

Handle `SYN_DROPPED` events in the evdev keyboard backend to prevent stale modifier key state.

When the kernel's evdev client buffer overflows, it inserts a `SYN_DROPPED` event to signal that events were lost ([kernel docs](https://www.kernel.org/doc/html/latest/input/event-codes.html)). Without handling this, a missed key-release event can leave `key_pressed_devices` in a stale state — for push-to-talk users this means recording may never stop or never start.

The fix discards all events between `SYN_DROPPED` and the next `SYN_REPORT`, then resets modifier key state for the affected device. The handling is per-device, so a drop on one keyboard does not affect another.

Reference: [libevdev SYN_DROPPED guide](https://www.freedesktop.org/software/libevdev/doc/latest/syn_dropped.html)

## Related Issue

Fixes #338

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Code refactoring
- [ ] Test update

## Checklist

- [x] Code follows project style (black, isort)
- [x] All tests pass locally
- [x] Pre-commit hooks pass

## Additional Notes

This is an edge case — keyboard events are sparse and the kernel buffer holds 64+ events, so overflow requires the reader thread to stall for multiple seconds. But the fix is ~15 lines, zero risk, and prevents a catastrophic stuck-key failure mode.

### DIFF
--- a/src/vocalinux/ui/keyboard_backends/evdev_backend.py
+++ b/src/vocalinux/ui/keyboard_backends/evdev_backend.py
@@ -160,6 +160,7 @@ class EvdevKeyboardBackend(KeyboardBackend):
         self.key_pressed_devices: set[int] = set()
 
         self._devices_lock = threading.Lock()
+        self._dropped_devices: set[int] = set()  # fds with SYN_DROPPED pending
 
         if not EVDEV_AVAILABLE:
             logger.error("python-evdev not available")
@@ -247,6 +248,7 @@ class EvdevKeyboardBackend(KeyboardBackend):
         self.devices = []
         self.device_fds = []
         self.key_pressed_devices = set()
+        self._dropped_devices = set()
 
         for device_path in device_paths:
             try:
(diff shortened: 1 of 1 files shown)


jatinkrmalik (approved):
@Sbenazar Great fix! The SYN_DROPPED handling is clean and well-explained. Thanks for the contribution!

Issue #302, opened 2026-03-13
[BUG]  No text typed after silence timeout


laurentpayot:
## 🐛 Bug Description
After I spoke, I wait for the `silence_timeout` seconds (and more) and nothing happens. I have to use the shortcut key to end the recording so my speech is typed.

Is it a bug or a feature?

## 📋 Steps to Reproduce
1. Press or hold shortcut key
2. Speak
4. Wait a few seconds in silence 
5. Nothing happens except if you press the shortcut key
## ✅ Expected Behavior
Speech should be processed *and typed* after the silence duration defined in the silence timeout setting.

## ❌ Actual Behavior
Nothing happens after the silence duration defined in the silence timeout setting.
The tray icon does not become orange and nothing is typed.

## 💻 Environment
- **OS**: Ubuntu 25.10
- **Python Version**: 3.13.7
- **Vocalinux Version**: 0.8.0-beta
- **Display Server**: Wayland
- **Speech Engine**: Whisper
- **Model Size**: small

## 📋 Debug Logs
Run `vocalinux --debug` and paste relevant log output:
```
2026-03-13 15:10:34,731 - vocalinux.single_instance - INFO - Single instance lock acquired
2026-03-13 15:10:34,733 - vocalinux.main - DEBUG - Debug logging enabled
2026-03-13 15:10:34,856 - vocalinux.utils.resource_manager - DEBUG - Checking resources candidate: /home/laurent/.local/share/vocalinux/venv/lib/python3.13/resources (exists: False)
2026-03-13 15:10:34,856 - vocalinux.utils.resource_manager - DEBUG - Checking resources candidate: /home/laurent/.local/share/vocalinux/venv/share/vocalinux/resources (exists: True)
2026-03-13 15:10:34,856 - vocalinux.utils.resource_manager - DEBUG - Checking resources candidate: /home/laurent/.local/share/vocalinux/resources (exists: False)
2026-03-13 15:10:34,856 - vocalinux.utils.resource_manager - DEBUG - Checking resources candidate: /usr/share/ubuntu/vocalinux/resources (exists: False)
2026-03-13 15:10:34,856 - vocalinux.utils.resource_manager - DEBUG - Checking resources candidate: /usr/share/gnome/vocalinux/resources (exists: False)
[...]

Issue #247, opened 2026-02-22
Whisper engine hallucinates repetitive nonsensical text after typing is stopped


christophtschmidt:
## 🐛 Bug Description
When using whsiper speech engine (medium). After "typing" some text via speech and stopping the vocie typing via "ctrl ctrl":  strange random text is generated after a while.

Whisper engines (both OpenAI Whisper and whisper.cpp) hallucinate repetitive
nonsensical text when processing silence or low-energy ambient noise. After a
successful transcription, background noise exceeding the VAD threshold fills
the audio buffer. On the next silence timeout this noise-only buffer is sent
to Whisper, which produces the same gibberish text repeatedly.

## 📋 Steps to Reproduce
1.  use whisper speech engine (medium for me)
2. start voice typing
3. entering text via speech
4. stop voice typing
5. wait a few seconds
6. random text is recognized and typed

## ✅ Expected Behavior
No hallucinated text appears

## ❌ Actual Behavior
random/hallucianted text is recognized and typed

## 💻 Environment
- **OS**: Fedora KDE 42
- **Python Version**: v0.6.3-beta
- **Vocalinux Version**: v0.6.3-beta
- **Display Server**:  Wayland
- **Speech Engine**: Whisper
- **Model Size**: medium 

I have created a patch that works for me.

Description:

Add multi-layered, language-agnostic anti-hallucination for Whisper and
whisper.cpp engines.
    
Whisper engines (both OpenAI Whisper and whisper.cpp) hallucinate repetitive
nonsensical text when processing silence or low-energy ambient noise. After a
successful transcription, background noise exceeding the VAD threshold fills
the audio buffer. On the next silence timeout this noise-only buffer is sent
to Whisper, which produces the same gibberish text repeatedly.

Layers of prevention added:
    
- Per-utterance speech tracking: only transcribe if VAD detected speech
- Audio energy gating: skip transcription when RMS < 150 (silence/noise)
- Minimum buffer duration: require >= 5 chunks (~300ms) of audio
- Structural hallucination detection: filter repetitive words (>60%) and
    repeated substrings (>60% coverage, 3+ occurrences)
- Duplicate transcription suppression: track recent results per session
- IDLE state guard: discard results that arrive after user stops

Add 4 new tests that exercise the real Whisper code path in
_process_final_buffer, catching regressions that were previously hidden
because all tests used the vosk engine or mocked the method entirely."

[...]

PR #351, opened 2026-03-26
feat(vad): replace amplitude-based VAD with Silero neural network

### LINKED ISSUE #344: Proposal: Replace amplitude-based VAD with Silero VAD
## Problem

The current VAD uses mean audio amplitude with a fixed threshold (`500 / sensitivity`). This causes two known issues:

- **#302** — silence doesn't stop recording because ambient noise exceeds the threshold
- **#247** — noise-filled buffers are sent to Whisper, producing hallucinated text

Amplitude-based VAD cannot reliably distinguish speech from background noise regardless of threshold tuning.

## Proposed Solution

Replace the amplitude VAD with [Silero VAD](https://github.com/snakers4/silero-vad) — a lightweight neural network model that classifies audio chunks as speech/non-speech.

**Key points:**

- Silero VAD runs via ONNX Runtime (no torch dependency required for inference), keeping the install lightweight
- The existing `vad_sensitivity` setting (1–5) would map to Silero's confidence threshold (e.g. 5 → 0.3, 1 → 0.8), preserving the current UI/config contract
- Fallback to the current amplitude VAD if Silero model is unavailable
- Minimal changes — the replacement is ~40 lines in `_record_audio()` plus model initialization

**Regarding #247:** Silero VAD addresses the root cause (noise reaching Whisper), while the hallucination token filter from [VocaMac PR #77](https://github.com/jatinkrmalik/vocamac/pull/77) would be a complementary defense layer — I'd propose that as a separate PR.

## Alternatives Considered

- **WebRTC VAD** — fast but less accurate than Silero, higher false-positive rate on background noise
- **Tuning amplitude threshold** — doesn't solve the fundamental problem; environment-dependent

Would love to hear your thoughts on this approach before writing the PR.

Sbenazar:
## Summary

- Replace the simple amplitude threshold VAD with [Silero VAD](https://github.com/snakers4/silero-vad) (v5), a lightweight neural network model that classifies audio as speech/non-speech
- Add `silero_vad.py` wrapper using ONNX Runtime with bundled `silero_vad.onnx` model (2.3MB)
- Accumulate 512-sample chunks with 64-sample context window per [official Silero API](https://github.com/snakers4/silero-vad/blob/master/src/silero_vad/utils_vad.py)
- Map existing `vad_sensitivity` (1–5) to Silero confidence thresholds (0.8–0.3)
- Fall back gracefully to the existing amplitude VAD when `onnxruntime` is unavailable
- Add `onnxruntime` as optional dependency (`pip install vocalinux[vad]`)

Addresses the root cause of #302 (silence not stopping recording) and #247 (noise-filled buffers causing Whisper hallucinations).

**Note on `silero-vad-lite`:** As discussed in #344, `silero-vad-lite` has no Python 3.14 wheels ([PyPI](https://pypi.org/project/silero-vad-lite/#files), [issue #7](https://github.com/daanzu/py-silero-vad-lite/issues/7)) and its bundled `.so` fails on glibc 2.41+ ([details in #344](https://github.com/jatinkrmalik/vocalinux/issues/344)). Using `onnxruntime` directly provides broader compatibility.

## Test plan

- [x] Silence produces low `silero_prob` values, no false speech detection
- [x] Speech correctly detected (`silero_prob=0.98-0.99`)
- [x] Silence timeout triggers segment flush after configured delay
- [x] Fallback to amplitude VAD when ONNX model is missing
- [x] `black` and `isort` pass
- [x] Tested on Fedora 43 / Wayland / AMD Radeon 780M

Closes #344


jatinkrmalik (changes requested):
Hey @Sbenazar, really nice contribution! The approach is solid and aligns well with our offline/privacy-first goals. Silero via ONNX is a smart choice over WebRTC VAD.

A few things to address before we can merge:

**Tests**
No tests included for the new silero_vad module or the modified VAD logic in _record_audio(). We need at least:
- Unit tests for SileroVAD.process() and reset()
- Test that load_silero_vad() returns None gracefully when onnxruntime is missing
- Test for the sensitivity-to-threshold mapping (1->0.8, 5->0.3)

**License attribution for the bundled model**
The Silero model is MIT licensed which is compatible with our GPLv3, but we should document that. Could you add a small LICENSE or NOTICE file in the data/ directory noting the Silero VAD license and version?

**Binary file in git**
The .onnx file is around 2MB which is manageable, but it lives in git history forever. I'm okay with keeping it bundled given the small size, but wanted to flag it. If you feel strongly about it we could also do a download-on-first-use pattern like we do for Whisper/VOSK models, but that's optional.

**Minor code note**
In _record_audio(), the speech_prob variable is initialized to 0.0 before the while loop that processes 512-sample chunks. If the first audio read doesn't fill a full 512-sample chunk, speech_prob stays at 0.0 and that iteration will always register as silence. Might be worth skipping the is_speech check until at least one full chunk has been processed.

Otherwise the implementation looks clean. The chunk accumulation, context window, and max() across sub-chunks all follow the Silero API correctly. Thanks for putting this together!

jatinkrmalik:
@Sbenazar I did a quick check on whether this is still useful after #415.

My read: yes, the PR is still valid and not redundant. #415 tunes whisper.cpp after audio reaches the decoder. This PR improves the upstream gate so silence/noise-only buffers are less likely to reach Whisper at all. Those are complementary fixes.

One user-facing issue: as written, users do not get a new setting for this. Silero VAD is only used automatically if `onnxruntime` is installed via the optional extra, e.g. `pip install "vocalinux[vad]"`. Normal installs will still fall back to amplitude VAD. The existing `vad_sensitivity` setting is reused and mapped to Silero thresholds, but users will not know which VAD backend is active.

Before merging, I would suggest:

1. Make sure the Silero license/notice is **included in packaged data**, not just committed in the repo.
2. Document how to enable neural VAD with `vocalinux[vad]`.
3. Make the active VAD backend visible somewhere, ideally in Settings near VAD sensitivity: “Voice detection: Silero neural VAD” or “Amplitude fallback”.
4. Consider a disabled/notice state when `onnxruntime` is missing, with the install command shown.
5. Rebase/update against current `main` and rerun CI.
6. Add one integration-style test for the `_record_audio()` Silero path, especially around silence timeout and push-to-talk behavior.

Bottom line: direction looks good, but the activation/discoverability story needs work. Otherwise most users may never actually get the new VAD.

Sbenazar:
Round-2 is up, rebased on current main. Walking through your six items:

1. LICENSE + onnx now ship in the wheel via package-data (verified locally w/ `python -m build`).
2. README has a short "Optional: Neural Voice Activity Detection" section w/ the `pip install "vocalinux[vad]"` command.
3. The VAD Sensitivity row subtitle in Settings shows the active backend ("Silero neural VAD" or "amplitude").
4. When onnxruntime is missing, an info-box appears under the VAD group w/ the install command and a restart hint -- same visual pattern as your Advanced-tab info box from #415.
5. CI now installs `.[dev,vad]` so onnxruntime is present in the test envrionment.
6. Added `tests/test_recognition_manager_silero.py` driving `_record_audio` w/ a mocked pyaudio stream + scripted SileroVAD probs -- covers silence flush, speech keeps the silence counter at 0, and the sens=1 vs sens=5 threshold split.

Also added a small `is_silero_available()` helper so the Settings UI doesn't pay the ~100ms ONNX session load just to render a label.

Let me know if the info-box wording needs tweaking.

Sbenazar:
Took the liberty of adding a few more tests to push patch coverage up -- it was bothering me to see codecov red while the rest of CI was green.

Added:
- `is_silero_available()` -- onnxruntime missing, model file missing, both present (silero_vad.py is at 100% now)
- amplitude fallback path in `_record_audio` (silero=None) -- silence flush + loud-audio-is-speech
- stereo capture (CHANNELS=2) and 48 kHz capture so the downmix and resample branches actually run
- audio level callback receives normalized values in [0, 100]

Local coverage on `vocalinux.speech_recognition` is now 85% (was 84% on main, 81% on the previous push of this branch).

One thing I want to flag on placement -- hope I put the UI bits in the right spot. The active-backend label is a subtitle on the VAD Sensitivity row, and the install-hint info-row sits right under the Recognition Settings group (borrowed the info-row visual from your #415 Advanced tab). If you'd rather see it as its own "Voice Activity Detection" section, or as a clickable button that opens a dialog like the whisper install one, just say the word and I'll move it -- not attached to the current layout.

src/vocalinux/speech_recognition/recognition_manager.py
```diff
@@ -2082,20 +2101,47 @@ def _record_audio(self):
...
+                            threshold = 500 / max(1, min(5, vad_sens))
+                        except ValueError:
+                            logger.warning(
+                                f"Invalid VAD sensitivity value: {self.vad_sensitivity}. Using default 3."
+                            )
+                            threshold = 500 / 3
+                        is_speech = volume >= threshold
 
-                    if volume < threshold:  # Silence
+                    if not is_speech:  # Silence
```
jatinkrmalik, line 2142: > `if not is_speech:  # Silence`

I noticed we still enqueue the buffer after `silence_timeout` even when every VAD decision for the segment was non-speech. For #444, a loopback or muted default device can keep producing silence-only buffers, so Whisper still gets a chance to hallucinate every timeout. Could we track something like `segment_has_speech` and drop/reset silence-only buffers, then only enqueue after speech followed by silence?

tests/test_recognition_manager_silero.py
```diff
@@ -0,0 +1,324 @@
...
+            language="en-us",
+            defer_download=True,
+        )
+    return mgr
+
+
+def _make_pyaudio_module(stream):
+    """Build a fake pyaudio module that returns the given stream from audio.open()."""
+    audio = MagicMock()
+    audio.get_device_count.return_value = 0
```
jatinkrmalik, line 75: > `audio.get_device_count.return_value = 0`

This makes `_resolve_valid_input_device(audio, None)` return `None`, so `_record_audio()` exits before the fake stream is read. That matches the current CI failures. Can we either return one fake input device with `maxInputChannels > 0`, or patch `_resolve_valid_input_device` to return `0` in these loop tests?

tests/test_recognition_manager_silero.py
```diff
@@ -0,0 +1,324 @@
...
+        """Silero state must be cleared between recording sessions."""
+        self._drive(probs=[0.1] * 5)
+        self.assertTrue(self.mgr._silero_vad.reset.called)
+
+    def test_silence_triggers_segment_enqueue(self):
+        """Sustained silence past silence_timeout should flush the audio buffer."""
+        # Each chunk = 1024/16000 = 0.064s. silence_timeout=0.5s, so ~8 chunks
+        # of silence flushes once. Run 20 chunks to ensure the threshold fires.
+        self._drive(probs=[0.05] * 20, vad_sensitivity=3)
+        self.assertGreaterEqual(
```
jatinkrmalik, line 187: > `self.assertGreaterEqual(len(self.enqueued), 1, ...)`

For the silence-only case, this assertion preserves the behavior that is hurting #444: silent input gets queued for transcription. Once we add a per-segment speech flag, this should probably become `assertEqual(len(self.enqueued), 0)` for all-silence input, with a separate speech-then-silence test proving real utterances still flush.

jatinkrmalik (commented):
Overall this is a good direction, but I would not merge yet. History check: this area has earlier fixes (`4aae8e8`, `6c80e31`, and #415) that filter or tune Whisper after audio reaches the decoder; this PR is complementary because it improves upstream VAD. Current blockers are failing Python CI and silence-only segments still being queued, which leaves #444 only partially fixed.

jatinkrmalik:
Hi @Sbenazar, thank you again for putting this together. This is a strong contribution and the Silero/ONNX direction is exactly the kind of offline-friendly fix we want here.

We just got a fresh report in #444 of Whisper hallucinating repeated text on silence, including with larger models, and this PR is the closest existing fix. I am going to take over the remaining production-hardening work so we can land it quickly.

The current blockers I want to fix on a maintainer branch are:

1. The Python CI jobs are red because the new `_record_audio()` tests never enter the mocked stream path.
2. The current recording loop still queues silence-only buffers after `silence_timeout`, which means a muted/loopback/default-silence source can still feed silent audio into Whisper. We need to track whether a segment actually had speech and drop pure-silence segments instead of decoding them.
3. I want the replacement PR to include regression coverage for the #444 behavior specifically: pure silence should not enqueue a transcription segment, while speech followed by silence still should.

I am going to close this PR and open a maintainer-owned replacement PR based on your branch/commits, with my fixes layered on top. You will still get attribution for the work:

- I will preserve your commits where possible by branching from this PR head.
- If the final merge is squashed, I will include `Co-authored-by: Sbenazar <127324689+Sbenazar@users.noreply.github.com>` in the squash commit message.
- The replacement PR description will explicitly credit you as the original contributor and link back to this PR.

Closing this only so we can move fast on the active user bug and get CI green from a branch I can maintain directly. Thank you for doing the hard part here.

jatinkrmalik:
Replacement maintainer PR is now open here: #447.

It is based on this PR's head so your commits remain in the branch history, and the PR body calls out the attribution trailer to preserve credit on squash merge:

```text
Co-authored-by: Sbenazar <127324689+Sbenazar@users.noreply.github.com>
```

PR #361, opened 2026-03-28
fix(ibus): start engine process before checking registration (fixes #360)

### LINKED ISSUE #360: [BUG] IBus engine setup fails — is_engine_registered() blocks start_engine_process() (follow-up to #303)
## Bug Description

`_setup_engine()` gates on `is_engine_registered()` which runs `ibus list-engine` to check if the Vocalinux engine is discoverable. On systems where the IBus daemon doesn't scan `~/.local/share/ibus/component/` (this directory is disabled in IBus source — `ibusregistry.c` has `#if 0` around user-dir scanning), the engine never appears in the list.

This causes `_setup_engine()` to raise `IBusSetupError` **before** `start_engine_process()` is ever called — even though `start_engine_process()` uses `register_component()` via D-Bus (added in PR #304), which is the correct registration path for standalone-launched engines.

A secondary issue: when the component XML is detected as outdated, `install_ibus_component()` triggers `ibus restart`, which kills a previously running engine process. This caused intermittent failures where the first dictation worked but subsequent ones failed with `Engine switch failed: expected vocalinux, got None`.

## Steps to Reproduce

1. Install Vocalinux on Fedora 43 / GNOME 49 / Wayland
2. Start `vocalinux-gui`
3. Observe log: `IBus initialization failed: Failed to register IBus engine. Try running 'ibus list-engine' to verify installation.`
4. Text injection falls back to ydotool (which garbles non-US/Cyrillic text)

## Expected Behavior

IBus engine should activate via `register_component()` D-Bus call, allowing Unicode text injection regardless of keyboard layout.

## Actual Behavior

- `is_engine_registered()` → `ibus list-engine` → engine not found → `IBusSetupError`
- Fallback to ydotool → Cyrillic text garbled (keycodes instead of Unicode)
- Even when IBus works initially, `install_ibus_component()` can trigger `ibus restart` and kill the engine process on subsequent launches

## Environment

- **OS:** Fedora 43, GNOME 49 / Wayland
- **Python:** 3.14
- **Vocalinux:** 0.10.0-beta
- **IBus:** system default (active input method)

## Root Cause

PR #304 correctly added `register_component()` support to the engine process, but `_setup_engine()` was not updated — it still gates on `ibus list-engine` which relies on the daemon's file-based component cache. The fix is to start the engine process first and rely on `register_component()` for runtime registration.

## Related Issues

- #303 / PR #304 — IBus engine activation timeout, direct predecessor of this fix
- #292 — keyboard layout reset on launch, caused by `ibus restart` in setup flow (which this PR removes)
- #266 — German umlauts truncated due to xdotool fallback (same root cause: IBus setup failure → non-Unicode fallback)

## Also Noticed

`wl-copy` clipboard copy times out (5s delay) after every text injection. This is a separate issue — `wl-copy` runs as a daemon process holding the clipboard, but Vocalinux waits for it to exit with `timeout=5`. Will file separately.

Sbenazar:
## Summary

- Remove `is_engine_registered()` gate from `_setup_engine()` — it queries `ibus list-engine` which fails on systems where IBus doesn't scan user component dirs (`~/.local/share/ibus/component/` is disabled in `ibusregistry.c`)
- Start engine process first, relying on `register_component()` D-Bus call for runtime registration (the correct path for standalone-launched engines, per PR #304)
- Remove `install_ibus_component()` from setup path — it triggers `ibus restart` which kills the engine process, causing intermittent `Engine switch failed` errors on subsequent dictation sessions
- Update class docstring and tests to match new setup flow

Follow-up to #303 / PR #304 which added `register_component()` support but left `_setup_engine()` gating on `ibus list-engine`.

## Test plan

- [x] IBus engine activates on first launch (Fedora 43 / GNOME 49 / Wayland)
- [x] Cyrillic text injection works via IBus `commit_text()`
- [x] Multiple dictation sessions with 2–3 minute pauses — engine stays alive
- [x] Fallback to ydotool/wtype still works when IBus is unavailable
- [x] `black` and `isort` pass
- [x] All existing tests pass (4 pre-existing Python 3.14 mock failures in `test_ibus_engine_core.py` unchanged)
- [x] Tested on Fedora 43 / Wayland / GNOME 49

Fixes #360


jatinkrmalik (changes requested):
👆 shared above

jatinkrmalik (approved):
Great work @Sbenazar! The fix is correct and well-reasoned as removing the `ibus list-engine` gate is the right call, and the socket-readiness poll is a solid improvement over the old approach.

Appreciate you addressing all three review points (dead code removal, readiness check, rebase) cleanly. 

One small follow-up: the `else` branch of the socket poll loop (the warning path when all 15 retries exhaust) has no test coverage as Codecov flagged it. 

I'll raise a follow-up PR to cover that so we keep the coverage baseline healthy. Thanks for the contribution!

Issue #94, opened 2026-01-29
Performance: Real-time Streaming Architecture for Lower Latency


jatinkrmalik:
## Problem Statement

Currently, Vocalinux processes audio in a pseudo-streaming manner with significant blocking operations that create a poor real-time user experience:

1. **Audio recording blocks** on PyAudio `stream.read()` calls
2. **Recognition processing blocks** - Whisper processes entire buffers at once, VOSK blocks during `AcceptWaveform()`
3. **Text injection has 100-500ms delays** with retry logic
4. **No proper inter-thread communication** - uses shared flags without locks (race condition risk)

This causes noticeable latency between speech and text appearing on screen.

## Proposed Solution

Implement a proper streaming pipeline architecture using producer-consumer pattern with queues and asyncio for I/O-bound operations.

### Architecture Overview

```
┌─────────────┐    ┌──────────┐    ┌──────┐    ┌─────────────┐    ┌──────────────┐
│ Audio Thread│───▶│ Audio Q  │───▶│ VAD  │───▶│ Recognition │───▶│ Text Inject  │
│ (Producer)  │    │          │    │      │    │ Thread      │    │ Queue        │
└─────────────┘    └──────────┘    └──────┘    └─────────────┘    └──────────────┘
                           │                              │
                           │                              ▼
                           │                         ┌─────────┐
                           └────────────────────────▶│ Callback│
                                                     │ Handler │
                                                     └─────────┘
```

### Phase 1: Queue-Based Audio Pipeline

**File: `src/vocalinux/speech_recognition/recognition_manager.py`**

1. **Add proper queue imports and constants** (around line 20):
```python
import asyncio
from queue import Queue, Empty
import concurrent.futures
from threading import Event, Lock

# Audio processing constants
AUDIO_QUEUE_MAX_SIZE = 50  # Prevent unbounded memory growth
AUDIO_CHUNK_DURATION = 0.064  # 1024 samples @ 16kHz = 64ms
```

2. **Replace the audio buffer with a thread-safe queue** (around line 120-130):
```python
# In RecognitionManager.__init__
self._audio_queue: Queue[bytes] = Queue(maxsize=AUDIO_QUEUE_MAX_SIZE)
self._queue_lock = Lock()
self._audio_event = Event()  # Signal when new audio is available
self._stop_event = Event()   # Signal for graceful shutdown
```

3. **Refactor audio recording to use queue** (around line 1016-1078 in `_record_audio`):
```python
[...]

PR #387, opened 2026-04-06
feat: experimental real-time streaming transcription (fixes #320)

### LINKED ISSUE #320: [BUG] Real Time Transcription doesn't work.

jatinkrmalik:
## Summary

Implements real-time streaming transcription as an **experimental feature** (off by default) that shows text as you speak, rather than waiting until you stop recording.

Fixes #320
Related discussions: #382, #94

### Phase 1: Config & Engine Plumbing
- **`config_manager.py`**: Added `experimental_streaming`, `streaming_chunk_duration_ms`, `streaming_overlap_ms` to DEFAULT_CONFIG (all off/default)
- **`common_types.py`**: Added `StreamingCallbackProtocol` and extended `SpeechRecognitionManagerProtocol` with streaming methods
- **`main.py`**: Passes streaming config from settings to `SpeechRecognitionManager` constructor

### Phase 2: TranscriptBuffer (LA-2 Dedup)
- **`transcript_buffer.py`**: New file implementing Local Agreement (LA-2) policy for word-level dedup across overlapping audio segments. Prevents flickering and duplicate text injection.

### Phase 3: Streaming Engine
- **`recognition_manager.py`**: 
  - Timer-based audio chunking with configurable overlap
  - Vosk path: uses native `PartialResult()` API
  - Whisper path: uses `TranscriptBuffer` with sliding window
  - `add_streaming_callback` / `remove_streaming_callback` for external consumers
  - `reconfigure` accepts streaming params with validation

### Phase 4: UI & Integration
- **`settings_dialog.py`**: Added streaming toggle and chunk duration spin in Recognition Settings
- **`tray_indicator.py`**: Registers streaming callback for live text injection

### Tests
- **`test_streaming.py`**: 23 unit tests covering TranscriptBuffer logic (basic, flush, reset, overlap, case-insensitive), config defaults, and config manager round-trip integration
- **`test_main.py`**: Updated for new constructor kwargs

## Architecture

```
┌─────────────┐    timer    ┌──────────────────┐    callback        ┌──────────────┐
│ Recording   │───enqueue──▶│ Streaming Engine │───(text,is_final)─▶│ TrayIndicator│
│ Loop        │  (overlap)  │ (Vosk/Whisper)   │                    │ → inject_text│
└─────────────┘             └──────────────────┘                    └──────────────┘
                                     │
                            ┌────────┴────────┐
                            │ TranscriptBuffer│
                            │ (LA-2 dedup)    │
                            └─────────────────┘
```

## Testing

- ✅ 23 new streaming tests pass
- ✅ 1455 existing tests pass (8 pre-existing failures in xkb_layout/ibus unrelated to this PR)
- ✅ All changed files formatted with `black --line-length 100`
- ✅ `flake8` clean on all changed files
- ✅ All files compile without syntax errors

## How to Test

1. Open Settings → Recognition Settings
2. Enable "Real-time Streaming" toggle
3. Start dictation — text should appear incrementally as you speak
4. Adjust "Chunk Duration" (0.2s–5.0s) to balance latency vs accuracy

## Notes

- Feature is **OFF by default** behind `experimental_streaming` config key
- Vosk streaming uses native `PartialResult()` — fast and well-tested
- Whisper streaming uses a sliding window with TranscriptBuffer dedup — experimental and may be less accurate
- The `TranscriptBuffer` LA-2 policy ensures words are only committed when seen in consecutive passes, preventing duplicates from overlapping audio segments


jatinkrmalik:
A note from local testing/investigation:

The streaming behavior is much more straightforward for Vosk than for Whisper/whisper.cpp. Vosk has a native incremental recognizer API: we can feed audio continuously via `AcceptWaveform()`, show `PartialResult()` as a live hypothesis, and commit `Result()` / `FinalResult()` when the recognizer endpoints. That model maps pretty naturally to dictation.

Whisper is different. It is fundamentally a windowed transcription model, not a true streaming decoder. For the experimental mode here we are repeatedly transcribing short overlapping chunks and then trying to decide what text is stable enough to inject. That creates a few practical problems:

- Very short chunks improve latency but make punctuation and phrase boundaries worse.
- Overlapping chunks often repeat boundary words, e.g. `are... are`, `of... of`, so we need tail/head deduping.
- Chunks can be sequential rather than revisions of the same phrase, so a strict Local Agreement policy can accidentally hold all text until stop.
- Injecting partial text directly into arbitrary apps is risky because replacing previous partials reliably across IBus/X11/Wayland/clipboard paths is a much bigger UI/input problem than just recognizing speech.

So my current read is: keeping this as an opt-in experimental mode is the right shape. Vosk streaming can be treated as closer to "native streaming". Whisper/whisper.cpp streaming should be framed as a best-effort low-latency chunking mode, where users may need to tune chunk duration upward for better punctuation and flow.

jatinkrmalik:
Closing since not viable for model families.

PR #389, opened 2026-04-06
fix(ibus): handle engine instance destruction on layout switch (fixes #388)

### LINKED ISSUE #388: [BUG] Text injection fails silently after keyboard layout switch (Super+Space)
## Description

When switching keyboard layout with Super+Space between dictation sessions, text injection fails silently -- the transcription works but nothing appears on screen.

## Steps to reproduce

1. Start Vocalinux with IBus text injection
2. Dictate some text (works fine)
3. Switch keyboard layout with Super+Space (e.g. en -> ru)
4. Switch back to en
5. Dictate again -- transcription runs but no text is inserted

## Root cause

When the user switches layout, IBus calls `do_destroy()` on the `VocalinuxEngine` instance. The engine process stays alive and the socket server keeps running, but `_active_instance` still points to the destroyed engine. The socket server sees `_active_instance` is set and tries to inject via GLib.idle_add, but the destroyed instance can't commit text.

The socket returns `OK` because `_active_instance` is non-None, but the actual `inject_text()` on the dead instance does nothing.

## Expected behavior

Text injection should recover automatically after a layout switch, either by detecting the stale instance or by retrying with a fresh engine.

## Environment

- Vocalinux v0.10.1-beta
- Fedora 43, GNOME 49 / Wayland
- IBus text injection mode
- Tested with en/ru layout switching

Sbenazar:
## Summary

Fixes #388. When switching keyboard layout (Super+Space), IBus calls `do_destroy()` on the engine instance but the process stays alive with a stale `_active_instance`. Subsequent dictations fail silently -- transcription runs fine but no text appears.

I hit this while testing #361 w/ ru/en layout switching between dictation sessions.

**Two changes:**

- **`VocalinuxEngine.do_destroy()`** -- clears `_active_instance` when IBus destroys the engine, so the socket server correctly returns `NO_ENGINE` instead of trying to inject via a dead instance
- **Retry in `IBusTextInjector.inject_text()`** -- on `NO_ENGINE` response, re-activates via `switch_engine()` and retries once. This creates a fresh engine instance and the second attempt succeeds

## Test plan

- [x] Added test: `test_do_destroy_clears_active_instance`
- [x] Added test: `test_do_destroy_ignores_different_instance`
- [x] Added test: `test_do_destroy_calls_super_when_ibus_available`
- [x] Added test: `test_inject_text_no_engine_retries_and_recovers`
- [x] Added test: `test_inject_text_no_engine_fails_after_retry`
- [x] Manual test: en<->ru layout switch between dictations, text injects correctly after switch
- [x] 117 tests pass (4 pre-existing failures on Python 3.14 unrelated to this change)

### DIFF
--- a/src/vocalinux/text_injection/ibus_engine.py
+++ b/src/vocalinux/text_injection/ibus_engine.py
@@ -565,6 +565,14 @@ class VocalinuxEngine(IBus.Engine if IBUS_AVAILABLE else object):
         # want to be able to inject text. The engine process keeps running and
         # the instance remains valid for text injection.
 
+    def do_destroy(self) -> None:
+        """Called when the engine instance is destroyed by IBus (e.g. layout switch)."""
+        logger.debug("VocalinuxEngine instance destroyed")
+        if VocalinuxEngine._active_instance is self:
+            VocalinuxEngine._active_instance = None
+        if IBUS_AVAILABLE:
+            super().do_destroy()
+
     def do_focus_in(self) -> None:
         """Called when the engine gains focus."""
         logger.debug("VocalinuxEngine focus in")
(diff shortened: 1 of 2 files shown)


jatinkrmalik:
Rebased onto latest main — the upstream drift (README, web, version files) is gone, now only `ibus_engine.py` and its tests remain. All 69 IBus tests pass locally. Waiting on CI.

PR #400, opened 2026-04-14
Add Moonshine STT support

Hieraphant:
## Summary

Adds Moonshine as a fourth speech recognition engine in Vocalinux.

This keeps the existing audio capture, buffering, and segmentation flow in place and adds Moonshine as another backend instead of changing the overall architecture.

## Changes

- add optional `moonshine-voice` dependency
- add `moonshine` to engine selection paths
- add config support for Moonshine model selection
- add Moonshine initialization and transcription in `recognition_manager`
- convert buffered PCM audio into the float sample format Moonshine expects
- add language mapping and fallback handling for unsupported languages
- add settings dialog support for Moonshine
- keep voice commands disabled by default for Moonshine in the settings path
- add focused tests for config/helper behavior

## Install

```bash
pip install "vocalinux[moonshine]"
```

Or:

```bash
pip install moonshine-voice
```

## Notes

This is a minimal v1 backend integration. It reuses the existing audio flow and does not introduce a new recognition pipeline.

Moonshine model selection is handled separately where needed instead of forcing a wider cross-engine model refactor.

## Local verification

Tested locally with `moonshine-voice 0.0.56`.

Verified locally:
- app starts with `--engine moonshine`
- Moonshine initializes successfully
- end-to-end transcription works through the existing Vocalinux flow
- a spoken test phrase was transcribed and injected successfully

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,6 +57,9 @@ whisper = [
     "torch>=2.0.0",
     "torchaudio>=2.0.0",
 ]
+moonshine = [
+    "moonshine-voice>=0.0.56",
+]
 dev = [
     "pytest>=7.0.0",
     "pytest-cov>=4.0.0",
--- a/src/vocalinux/main.py
+++ b/src/vocalinux/main.py
@@ -61,9 +61,7 @@ def parse_arguments():
         choices=["vosk", "whisper", "whisper_cpp", "moonshine"],
         help="Speech recognition engine to use (whisper_cpp recommended for best performance)",
     )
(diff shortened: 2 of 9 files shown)


Hieraphant:
Follow-up updates are now pushed.

Changes since the initial PR:
- fixed the config-manager test to include the new `moonshine` engine
- added targeted tests for the Moonshine recognition-manager path
- fixed the Moonshine settings UI so it no longer shows fake `0 MB` model sizes
- verified the local test suite passes after the updates

PR #401, opened 2026-04-16
fix(recognition): preserve final speech on stop

sabbari:
## Summary
- replace the hard-coded stop-time tail trim with a configurable `stop_sound_guard_ms`
- preserve short final utterances that were previously clipped on key release
- add focused test coverage for the stop-recognition path

## Motivation
In push-to-talk use, the final spoken word could be lost unless the user waited briefly before releasing the key. The root cause was a hard-coded trim of roughly the last second of recorded audio during stop, which could remove real speech instead of only suppressing the stop sound.

## Notes
- default guard is `200ms`
- this keeps the stop-sound protection while avoiding the previous speech cutoff

### DIFF
--- a/src/vocalinux/main.py
+++ b/src/vocalinux/main.py
@@ -347,6 +347,7 @@ def main():
 
     vad_sensitivity = saved_settings.get("vad_sensitivity", 3)
     silence_timeout = saved_settings.get("silence_timeout", 2.0)
+    stop_sound_guard_ms = saved_settings.get("stop_sound_guard_ms", 200)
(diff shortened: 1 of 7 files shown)


jatinkrmalik:
Took over the test fix — existing tests were asserting `SpeechRecognitionManager` was called without `stop_sound_guard_ms`, but the PR adds it as a new kwarg. Updated `test_main_initializes_components` to expect `stop_sound_guard_ms=200`. Pushed to branch, waiting for CI.

PR #402, opened 2026-04-16
feat(injection): add selectable text injection backends

sabbari:
## Summary
- add a configurable text injection backend setting in the GTK settings dialog
- expose the existing backend choices: `auto`, `ibus`, `xdotool`, `wtype`, and `ydotool`
- apply backend changes immediately by rebuilding the active injector at runtime

## Motivation
On X11, text injection could still route through IBus whenever it initialized successfully, even in applications where IBus was not reliable in practice, such as VS Code and some terminal contexts. This made injection behavior hard to control and hard to debug, especially because an accepted IBus request did not necessarily mean that text was actually inserted in the focused application.

## Notes
- explicit backend selections now fail clearly when they are incompatible with the current session type
- this is especially useful on X11 setups where `xdotool` works better than IBus in some apps

### DIFF
--- a/src/vocalinux/main.py
+++ b/src/vocalinux/main.py
@@ -288,10 +288,12 @@ def main():
     logger.info("Logging system initialized")
 
     config_manager = ConfigManager()
-    saved_settings = config_manager.get_settings().get("speech_recognition", {})
-    audio_settings = config_manager.get_settings().get("audio", {})
+    all_settings = config_manager.get_settings()
+    saved_settings = all_settings.get("speech_recognition", {})
+    audio_settings = all_settings.get("audio", {})
+    text_injection_settings = all_settings.get("text_injection", {})
 
-    general_settings = config_manager.get_settings().get("general", {})
+    general_settings = all_settings.get("general", {})
     first_run = general_settings.get("first_run", True)
     should_prompt_first_run = first_run and not args.start_minimized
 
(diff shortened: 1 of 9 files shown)


jatinkrmalik:
@sabbari Thanks for this PR, I think this is useful and I’d like to get it in before the next release. I don’t have push access to your fork, so could you please rebase this branch on the latest `main` and resolve the merge conflicts? 

Once it’s updated, I’ll prioritize the review and merge.

jatinkrmalik:
Hey @sabbari, closing this in favor of what's already on `main` plus #649.

I asked for a rebase in May so we could get this into a release, but the branch hasn't moved. It's conflicting, lint has been red since April (black), and the test matrix never ran.

`main` already has `VOCALINUX_FORCE_BACKEND` (`ibus` / `wtype` / `ydotool` / `auto`) in `TextInjector._forced_backend()`. #649 is the persistent `text_injection.backend` setting, env-overrides-config, default `auto`. That's the narrower version of this PR (no Settings combo, no runtime rebuild).

The extra work here also doesn't switch dictation: `update_text_injection_backend` replaces `tray.text_injector`, but `main()` still holds the old injector on `text_system` / `ActionHandler`. And `_configure_ibus()` uses `IBusTextInjector(auto_activate=True)`, which we'd have to rewrite against scoped injection (`auto_activate=False`) anyway.

The one thing #649 left out on purpose is pinning `xdotool`. If that's still needed (X11 #476, Firefox IBus), let's add it on #649. Please don't rebase this branch. Thanks for the original work.

PR #406, opened 2026-04-24
feat(settings): expose initial_prompt for whisper.cpp and Whisper 🤖🤖🤖

### LINKED ISSUE #393: [FEATURE] Expose whisper.cpp initial_prompt in config and settings UI`
## 🚀 Feature Description

Expose the whisper.cpp `initial_prompt` parameter in
`config.json` and the Settings UI, allowing users to seed
the transcription model with custom vocabulary, spelling
conventions, and domain-specific terms.

## 💡 Problem Statement

Users have no way to influence transcription style or correct
consistent errors without patching source code.

- British/Australian English speakers get American spellings
  ("organize" instead of "organise", "color" instead of
  "colour") with no way to override.
- Technical jargon, project names, and acronyms are
  consistently mistranscribed (e.g. "i3wm", "sxhkd",
  "PipeWire").
- Users with accents or speech patterns that cause repeated
  misrecognitions have no corrective mechanism.

## 📋 Proposed Solution

1. Add an `initial_prompt` field to the `speech_recognition`
   section of `config.json`:

   ```json
   "speech_recognition": {
       "engine": "whisper_cpp",
       "initial_prompt": "British English: organisation, colour, favourite. Terms: Vocalinux, Kubernetes, GitLab.",
   }
   ```

2. Add a multi-line text area to **Settings → Recognition**
   for editing the prompt.

3. Pass the value through to the whisper.cpp `Model()`
   constructor in `recognition_manager.py`. `pywhispercpp`
   already accepts `initial_prompt` as a kwarg — this is
   primarily a config plumbing change.

## 🔄 Alternatives Considered

- **Post-processing regex replacements:** Could auto-correct
  specific words after transcription, but doesn't influence
  the model's token probabilities so accuracy remains poor.
- **Per-language model fine-tuning:** Far too heavy for
  individual vocabulary tweaks.
- **External wrapper script:** Currently using a workaround
  that calls `pwcpp` directly with `--initial_prompt` for
  terminal dictation, but this bypasses Vocalinux entirely.

## 📸 Mockups / Examples

The prompt would appear as a text area in the Recognition
settings tab, below the existing VAD Sensitivity and Silence
Timeout controls:

```
┌─ Recognition Settings ─────────────────────┐
│ VAD Sensitivity          [3] [-] [+]        │
│ Silence Timeout          [2.0] [-] [+]      │
│ Voice Commands           [toggle]            │
│                                              │
│ Initial Prompt                               │
│ ┌──────────────────────────────────────────┐ │
│ │ British English: organisation, colour,   │ │
│ │ favourite. Terms: Vocalinux, Kubernetes. │ │
│ └──────────────────────────────────────────┘ │
│ Hint: Bias transcription toward specific     │
│ vocabulary, spelling, or style.              │
└──────────────────────────────────────────────┘
```

## 📝 Additional Context

- The `initial_prompt` parameter is supported by both
  whisper.cpp and OpenAI Whisper engines. VOSK does not
  support it, so the field should be ignored/hidden when
  VOSK is selected.
- Very long prompts can degrade performance — a character
  limit or warning in the UI would be helpful.
- This is a well-established Whisper feature used by
  SuperWhisper (macOS) and other voice typing tools.

## ✅ Acceptance Criteria

- [ ] `initial_prompt` field added to `config.json` under
      `speech_recognition`
- [ ] Text area added to Settings → Recognition tab
- [ ] Prompt passed to whisper.cpp and Whisper engines
      during transcription
- [ ] Field ignored/hidden when VOSK engine is selected
- [ ] Empty prompt has no effect on existing behaviour
- [ ] Prompt persists across restarts

almendio-dev:
## Summary

Add support for seeding transcription models with custom vocabulary, spelling conventions, and domain-specific terms via config.json and the Settings UI.

## Changes

- Add `initial_prompt` to config.json (default: empty string)
- Add GTK UI section in Recognition Settings with:
  - Debounced save (3s idle) + immediate save on focus-out
  - 500-character limit with warning styling at >100 chars
  - Hidden when VOSK engine is selected
- Pass prompt to whisper.cpp Model() constructor
- Pass prompt to OpenAI Whisper transcribe() method
- Omit empty prompt to avoid behavior change
- Add try/except TypeError fallback for older pywhispercpp
- Trigger model reload on whisper.cpp when prompt changes
- Add tests for config round-trip and reconfigure reload logic

## Testing

- [x] Code follows style guidelines (black, isort, flake8 pass)
- [x] Tests pass locally (except environment-specific tests unrelated to changes)
- [x] Commit messages follow Conventional Commits

## Related Issues

Fixes #393


jatinkrmalik:
@almendio-dev Thanks for the contribution! The core logic is correct and the UX intent is solid. Please address the following before merge:

---

## Blockers / Major

**1. GPU fallback crashes on old pywhispercpp (regression)**

`_handle_gpu_fallback` passes `model_kwargs` -- which still contains `initial_prompt` -- to `Model()` without any TypeError protection. On a system with old pywhispercpp + a Vulkan failure, the CPU retry raises an unhandled `TypeError` and the user can't load any model at all.

Recommended fix: detect support once upfront with `inspect.signature` instead of catching exceptions:

```python
import inspect
if "initial_prompt" not in inspect.signature(Model.__init__).parameters:
    model_kwargs.pop("initial_prompt", None)
    logger.warning("pywhispercpp does not support initial_prompt (old version).")
```

This applies cleanly to both the primary load and the CPU fallback path without duplicating the try/except.

---

**2. `_on_dialog_response` only flushes when a debounce is pending**

```python
def _on_dialog_response(self, dialog, response_id):
    if self._prompt_debounce_id is not None:   # flush is inside the if
        GLib.source_remove(...)
        self._apply_prompt_setting()
```

If the debounce fires just before the dialog closes and resets `_prompt_debounce_id = None`, the user's last edit is silently dropped. `_apply_prompt_setting` already short-circuits when `text == current_prompt`, so it is safe and cheap to call unconditionally:

```python
def _on_dialog_response(self, dialog, response_id):
    if self._prompt_debounce_id is not None:
        GLib.source_remove(self._prompt_debounce_id)
        self._prompt_debounce_id = None
    self._apply_prompt_setting()  # always flush
```

---

**3. UI tests are grep over source, not behavioral tests**

Every test in `TestSettingsDialogInitialPrompt` does:

```python
self.assertIn("initial_prompt_textview", source_code)
```

These pass even if the handlers crash or the debounce never fires. They give a false coverage signal. Please replace with real behavioral tests (mock the dialog, drive `buffer.set_text()`, advance `GLib.MainContext.iteration()`, assert `config_manager.set` was called). Minimum cases to cover:

- Buffer > 500 chars is truncated
- `count > 100` adds `status-warning`; `count <= 100` removes it
- Focus-out cancels debounce and calls apply once
- `_apply_prompt_setting` no-ops when text equals stored prompt
- `_apply_prompt_setting` calls `speech_engine.reconfigure(initial_prompt=...)`
- `_update_initial_prompt_visibility` hides section for Vosk, shows it for Whisper / whisper_cpp

---

**4. `save_settings()` is deprecated -- use `save_config()`**

`_apply_prompt_setting` calls `self.config_manager.save_settings()`, which emits a `DeprecationWarning`. Every keystroke that triggers the debounce will spam the logs. Replace with `self.config_manager.save_config()` -- a one-liner fix.

---

**5. `codecov.yml` is unrelated scope**

This file changes coverage gate thresholds repo-wide and has no connection to `initial_prompt`. It is also missing a trailing newline. Please split it into a separate PR.

---

## Minor

**6. 100-char warning threshold is misleading**

Whisper's actual prompt token budget is ~224 tokens (~800-1000 English chars). Warning at 100/500 will make users think 101 chars is dangerous. Consider warning at 400 (near the real limit) or removing the `status-warning` class entirely and letting the counter speak for itself.

**7. Add a comment explaining the 3-second debounce**

The duration is correct (whisper.cpp requires a model reload on prompt change, so debouncing is important), but it looks arbitrary without explanation. A short comment goes a long way for future contributors.

**8. `reconfigure()` docstring is stale**

It lists all existing kwargs but not `initial_prompt`. Please add it.

**9. README JSON indentation is inconsistent**

The diff introduces 8-space indentation for inner keys while the outer object stays at 2-space. Valid JSON, but inconsistent with the surrounding style. Please revert the inner keys to 2-space.

---

## Nits

- `prompt_subtitle` and `initial_prompt_hint` convey nearly the same message -- two labels under one input is noisy; consider removing one.
- CSS comment reads `Use test-textview on non-editable TextViews only.` -- should be `.test-textview`.
- `test_settings_dialog_has_500_char_limit` asserts `"500" in source_code` -- "500" appears in many unrelated places (timeouts, heights) and is not meaningful even as a string check.
- No success-level log after `reconfigure()` returns; the current `logger.info("Applying initial_prompt...")` fires before the call.
- Worth noting in the tooltip or docs that prompts are language-specific -- a French prompt will bias an English transcription in unexpected ways.

---

Overall this is a well-structured addition and close to mergeable. The GPU fallback regression (item 1) and deprecated `save_settings` call (item 4) are the most important to fix before landing.

almendio-dev:
Thanks for your detailed review @jatinkrmalik - I  addressed them:


**Blockers / Major**

**1. GPU fallback crashes on old pywhispercpp** — Fixed. Replaced the `try/except TypeError` with an upfront `inspect.signature(Model.__init__)` check that detects support once, before `model_kwargs` is built. This protects both the primary load and the `_handle_gpu_fallback` CPU retry path without duplicating exception handling.

**2. `_on_dialog_response` only flushes when debounce is pending** — Fixed. Moved `self._apply_prompt_setting()` outside the `if self._prompt_debounce_id is not None:` block so it always runs on dialog close. Since `_apply_prompt_setting` already short-circuits when `text == current_prompt`, this is safe and cheap.

**3. UI tests are grep over source, not behavioral** — Fixed. Replaced the entire `TestSettingsDialogInitialPrompt` class with `TestInitialPromptHandlers` containing 13 behavioral tests that exercise the actual handler logic via mocks: config save + engine reconfigure on prompt change, no-op when text matches stored prompt, no-op during initialization, recursive-apply guard, graceful reconfigure failure handling, 500-char truncation, 400-char warning threshold, visibility per engine, and focus-out debounce cancellation.

**4. `save_settings()` is deprecated** — Fixed. Replaced with `save_config()` in `_apply_prompt_setting`.

**5. `codecov.yml` is unrelated scope** — Removed from this PR entirely. For the record: without it, project coverage is 80.18% — above the original 78.15% base.

**Minor**

**6. 100-char warning threshold** — Raised to 400. Agree that 100 was misleadingly low given Whisper's ~224-token budget.

**7. Debounce comment** — Added: "whisper.cpp requires a full model reload on prompt change, so we batch rapid edits to avoid reloading the model on every keystroke."

**8. `reconfigure()` docstring** — Updated to document `initial_prompt` (str) and `voice_commands_enabled` (bool) under a new `Keyword Args:` section.

**9. README JSON indentation** — Fixed to consistent 2-space.

**Nits**

- Kept both `prompt_subtitle` and `initial_prompt_hint` for now — the subtitle describes the field, the hint explains the effect. Happy to collapse if you prefer.
- Fixed the CSS comment to read `.test-textview`.
- Replaced `test_settings_dialog_has_500_char_limit` with proper truncation tests in the new behavioral suite.
- Noted on the tooltip — will leave as-is for this PR since it's already fairly detailed.

jatinkrmalik:
@almendio-dev I have actually incorporated some of your changes here: https://github.com/jatinkrmalik/vocalinux/pull/415 

I will attribute it to you in the next release. Once I re-merge it, you may want to re-base and see if there is anything else you would want to add. I have added a screencast demo as well in the pull request, so please take a look.

Thank you for the idea. I ended up exposing a lot more other Tweaks and knobs for whisper.cpp under the advanced tab

almendio-dev:
@jatinkrmalik I just saw your comment, thanks for the heads up! Really appreciate you incorporating the initial_prompt work and the attribution.

I'll take a look at #415, the screencast demo is a nice touch. The advanced tab with extra whisper.cpp knobs sounds great, looking forward to test what you've added there.

Thanks for the collaboration!

PR #413, opened 2026-05-03
fix(logging): clean up runtime log noise and cache hardware detection

jatinkrmalik:
## Summary

Found four issues while investigating a runtime log from a fresh install of latest `main` (post-v0.10.2 merges). Three are duplicated-code artifacts from recent merges; one is a perf bug that adds ~1.3 s of UI lag whenever the settings dialog opens.

### 1. Duplicate `_perform_recognition` loop body (`recognition_manager.py`)

Commit d049473 (#401, *preserve final speech on stop*) added a new `while True:` loop body but left the previous `while self.should_record or not self._segment_queue.empty():` body glued in immediately after, with a stray docstring acting as an expression statement. The dead loop ran once after every recording and produced a spurious extra:
```
DEBUG: _perform_recognition thread started
DEBUG: _perform_recognition thread exiting
```
visible in the attached log. Removed.

### 2. Duplicate startup init in `main.py`

Commit cb9b66a (#261) introduced a duplicated block:
```python
config_manager = ConfigManager()
initialize_logging()
logger.info(\"Logging system initialized\")

config_manager = ConfigManager()
```
Causing `Logging system initialized` and `Loaded configuration from /home/jk/.config/vocalinux/config.json` to log twice on every startup and reading the config file from disk once more than necessary. Removed the duplicate.

### 3. Hardware detection re-shells out on every settings-dialog interaction (`whispercpp_model_info.py`)

`detect_vulkan_support()`, `detect_cuda_support()`, `detect_cpu_info()`, and `detect_compute_backend()` shell out to `vulkaninfo --summary` / `nvidia-smi` every call. The settings dialog calls them while populating model dropdowns; the attached log shows **10+ Vulkan probes within ~1.3 s** every time the dialog opens.

Decorated all four with `@lru_cache(maxsize=1)` — hardware doesn't change at runtime. Added an autouse `conftest.py` fixture to clear the caches between tests so existing `subprocess.run` mocks keep working as written.

### 4. Demoted `logger.info(\"DEBUG: ...\")` breadcrumbs to `logger.debug(...)`

13 internal queue/thread breadcrumbs in `recognition_manager.py` were logged at INFO level with a literal `DEBUG:` prefix, flooding default-level logs. Demoted to `logger.debug` so they only appear with `--debug`.

## Verification

- `make lint` (black, isort, flake8) ✅
- Targeted tests: `test_whispercpp_model_info_ext.py`, `test_recognition_manager_*`, `test_whisper_support.py`, `test_misc_edge_cases.py` — **151 passed**
- Full suite (excluding `test_main*.py` which fail pre-existing on this machine because a Vocalinux instance is currently running and holds the lock — verified identical failures on clean `main`) — **1428 passed**

## Before vs After (representative log lines per recording)

Before (from the user-reported log):
```
[19:36:13.282] INFO ... DEBUG: _perform_recognition thread started
[19:36:13.282] INFO ... Loaded configuration from .../config.json
...
[19:36:16.371] INFO ... DEBUG: invoking text callback: text_callback_wrapper
[19:36:16.371] INFO ... DEBUG: _perform_recognition thread exiting
[19:36:16.371] INFO ... DEBUG: _perform_recognition thread started   ← dead loop
[19:36:16.371] INFO ... DEBUG: _perform_recognition thread exiting   ← dead loop
```

After: dead loop gone, all `DEBUG:`-prefixed lines moved to `logger.debug` (suppressed unless `--debug`), settings dialog no longer log-spams Vulkan detection.

### DIFF
--- a/src/vocalinux/main.py
+++ b/src/vocalinux/main.py
@@ -284,10 +284,6 @@ def main():
         logger.debug(f"Could not start IBus daemon: {e}")
 
     config_manager = ConfigManager()
-    initialize_logging()
-    logger.info("Logging system initialized")
-
-    config_manager = ConfigManager()
     saved_settings = config_manager.get_settings().get("speech_recognition", {})
     audio_settings = config_manager.get_settings().get("audio", {})
 
--- a/src/vocalinux/speech_recognition/recognition_manager.py
+++ b/src/vocalinux/speech_recognition/recognition_manager.py
@@ -1600,7 +1600,7 @@ class SpeechRecognitionManager:
                 )
 
             if self.audio_buffer:
-                logger.info(f"DEBUG: Enqueuing final buffer with {len(self.audio_buffer)} chunks")
+                logger.debug(f"Enqueuing final buffer with {len(self.audio_buffer)} chunks")
                 self._enqueue_audio_segment(self.audio_buffer)
                 self.audio_buffer = []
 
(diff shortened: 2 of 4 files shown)


jatinkrmalik:
Merged. All functional CI green — codecov/patch drop is from deduplicating code, not new untested logic.

PR #425, opened 2026-05-08
fix: paste into Firefox through clipboard fallback

juanfradb:
## Summary
- detect focused Firefox windows when using the IBus text injection path
- use clipboard copy plus `xdotool key ctrl+v` for Firefox instead of direct IBus commit
- keep the existing IBus path for other applications
- add tests for Firefox fallback, non-Firefox behavior, and X11 paste injection

## Why
Firefox text fields can ignore IBus commit text in some X11 desktop setups even while terminal and other applications work. Clipboard paste is the reliable path there and already matches the existing clipboard fallback model.

## Impact
Firefox users get the dictated text inserted instead of silently losing it. The recognized text is copied to clipboard as part of the paste fallback.

## Validation
- `PYTHONPATH=src /home/juanfra/.local/share/vocalinux/venv/bin/python -m pytest tests/test_text_injector.py -q`
- `PYTHONPATH=src /home/juanfra/.local/share/vocalinux/venv/bin/python -m py_compile src/vocalinux/text_injection/text_injector.py tests/test_text_injector.py`
- `black --check` on changed files

### DIFF
--- a/src/vocalinux/text_injection/text_injector.py
+++ b/src/vocalinux/text_injection/text_injector.py
@@ -534,6 +534,89 @@ class TextInjector:
         except Exception as e:
             logger.debug(f"Could not show clipboard notification: {e}")
 
+    def _get_current_x11_app_id(self) -> str:
+        """Return a best-effort identifier for the focused X11/XWayland application."""
+        if not shutil.which("xdotool"):
+            return ""
+
+        try:
+            env = os.environ.copy()
+            window = subprocess.run(
+                ["xdotool", "getactivewindow"],
+                env=env,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.DEVNULL,
+                text=True,
+                check=False,
+                timeout=1,
+            )
+            window_id = window.stdout.strip()
+            if window.returncode != 0 or not window_id:
+                return ""
+
+            app_parts = []
+            for args in (
+                ["xdotool", "getwindowclassname", window_id],
+                ["xdotool", "getwindowname", window_id],
+            ):
+                result = subprocess.run(
+                    args,
+                    env=env,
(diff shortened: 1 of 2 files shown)


jatinkrmalik:
Hey @juanfradb, closing this. Thanks for the Firefox report, but I don't want this patch to land.

#665 did not fix Firefox ignoring IBus. That PR restores the XKB layout after scoped injection. `IBusTextInjector.inject_text()` still returns `True` as soon as `commit_text()` succeeds at the IBus layer, so `_switch_to_non_ibus_backend()` never runs when Firefox drops the text. That failure mode can still be real. This isn't the fix for it.

A few problems with the approach:

1. `"firefox" in target_app` after joining window title + class + `/proc/cmdline`. Any focused window whose title mentions Firefox (Chrome, VS Code, a terminal) would skip IBus and paste. Detect `WM_CLASS` / `comm` only.
2. The same branch runs on `WAYLAND_IBUS` and then calls `xdotool key ctrl+v`, which does not inject into native Wayland Firefox.
3. `_inject_via_x11_clipboard_paste` reimplements paste and overwrites the clipboard with no restore. `main` already has `_inject_via_clipboard_paste` (restore + generation counter).

Draft, untouched since May, and it conflicts with current `inject_text()`.

If Firefox-on-X11 still silently drops IBus commits, the right knobs are already in flight: pin a non-IBus backend (`VOCALINUX_FORCE_BACKEND` on `main`; config pin in #649). Neither accepts `xdotool` today, which is the actual X11 gap. Add `xdotool` there, or denylist WM_CLASS `firefox` / `Navigator` and reuse `_inject_via_clipboard_paste`. Don't special-case by window title.

PR #426, opened 2026-05-08
fix: play stop sound immediately on release

juanfradb:
## Summary
- play the stop sound immediately after recording is marked stopped
- keep the existing stop-sound guard trim so the cue is not transcribed
- add a regression test that the cue plays before waiting for the audio thread to join

## Why
Push-to-talk users get delayed feedback today because the stop sound is played only after audio-thread shutdown and final buffer handling. Recording is already marked stopped first, and the existing guard trim handles any in-flight chunk, so the feedback can be moved earlier without changing transcription flow.

## Impact
The release cue feels immediate while preserving the current final-buffer transcription path.

## Validation
- `PYTHONPATH=src /home/juanfra/.local/share/vocalinux/venv/bin/python -m pytest tests/test_recognition_manager_advanced.py tests/test_recognition_manager_core.py -q`
- `PYTHONPATH=src /home/juanfra/.local/share/vocalinux/venv/bin/python -m py_compile src/vocalinux/speech_recognition/recognition_manager.py tests/test_recognition_manager_advanced.py`
- `black --check` on changed files

### DIFF
--- a/src/vocalinux/speech_recognition/recognition_manager.py
+++ b/src/vocalinux/speech_recognition/recognition_manager.py
@@ -1651,6 +1651,10 @@ class SpeechRecognitionManager:
         # Stop recording FIRST to prevent capturing the stop sound
         self.should_record = False
 
+        # Give immediate release feedback. The recorder may still finish one
+        # in-flight chunk, so the stop-sound guard below keeps the cue out.
+        play_stop_sound()
+
         # Wait for audio thread to finish recording and enqueue any pending audio
         # This is critical to prevent race condition where recognition thread exits
         # before the final audio segment is enqueued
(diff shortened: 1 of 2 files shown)


jatinkrmalik:
@juanfradb this is an interesting one. Is it ready for review? I see the PR is marked as draft.

jatinkrmalik:
Thanks for this contribution, @juanfradb, the intent to give push-to-talk users faster auditory feedback is a nice UX improvement.

I reviewed the change and noticed one gap for short recordings (≤ ~192 ms). In the current PR, playing the stop sound before `audio_thread.join()` means the audio thread can still capture that cue in its in-flight `stream.read()` chunk. The guard trim only runs when `len(audio_buffer) > stop_sound_guard_chunks`, so very short buffers aren't trimmed and the cue can leak into transcription.

I'm going to merge this PR as-is to keep momentum, and I'll follow up immediately with a small fix that moves `play_stop_sound()` to just after `audio_thread.join()` (so the thread is dead and can't capture it) but still before the final buffer processing so preserving the speed gain while closing the capture window.

PR #428, opened 2026-05-11
fix(flatpak): use 'flatpak run' for autostart Exec command inside sandbox

jatinkrmalik:
## Summary

Inside the Flatpak sandbox the `vocalinux` binary is not on the **host** PATH, so the XDG autostart desktop entry written by `autostart_manager.enable_autostart()` would silently fail — the session manager can't find `vocalinux` after login.

## Root Cause

`get_exec_command()` tried `shutil.which("vocalinux")` first. Inside the sandbox that resolves to `/app/bin/vocalinux`, which is inaccessible from the host login session.

## Fix

`get_exec_command()` now checks `FLATPAK_ID` first. When running inside a Flatpak sandbox it returns:

```
flatpak run com.vocalinux.Vocalinux --start-minimized
```

The host-side `flatpak` launcher is always on PATH (it's a system package), so autostart works correctly. Non-Flatpak install paths are unchanged.

## Tests

Two new tests added to `test_autostart_manager.py`:
- `test_get_exec_command_flatpak_uses_flatpak_run` — verifies the correct command is returned when `FLATPAK_ID` is set
- `test_enable_autostart_flatpak_writes_flatpak_run_exec` — verifies the generated `.desktop` file contains the correct `Exec=` line

All 17 autostart tests pass.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -140,6 +140,30 @@ jobs:
         run: |
           twine upload dist/* --skip-existing
 
+  build-flatpak:
+    name: Build Flatpak
+    needs: build-and-release
+    runs-on: ubuntu-latest
+    container:
+      image: bilelmoussaoui/flatpak-github-actions:gnome-46
+      options: --privileged
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Build Flatpak
+        uses: flatpak/flatpak-github-actions/flatpak-builder@v6
+        with:
+          bundle: vocalinux.flatpak
+          manifest-path: packaging/flatpak/com.vocalinux.Vocalinux.yml
+          cache-key: flatpak-${{ github.sha }}
+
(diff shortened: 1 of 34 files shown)


jatinkrmalik:
Closing this in favor of a fresh, scoped-down Flatpak implementation.

This branch (and #416) mixed the genuinely Flatpak-related changes with a number of unrelated runtime changes — whisper.cpp GPU kwargs, audio-device revalidation, clipboard timeout tuning, `wl-copy` rewrites — which made the diff noisy and hard to land cleanly. The new effort keeps only the surgical changes Flatpak actually needs (XDG base-dir paths, `flatpak run` autostart, tray icon naming, XWayland text injection) plus the packaging files, and is being validated with a real `flatpak-builder` build.

New branch: `feat/flatpak-packaging`. Thanks for the groundwork here — the manifest and shared-modules were a useful reference.

PR #437, opened 2026-05-30
fix(installer): harden Debian compatibility layer

jatinkrmalik:
## Summary

Fixes multiple Debian compatibility gaps that cause "installed but does not run" failures, documented in detail by a Debian user who had to manually work through CMake bootstrap errors, missing shared libraries, and missing text injection tools.

Reported in: https://medium.com/@cslev/talking-to-my-linux-box-without-talking-to-the-cloud-vocalinux-on-debian-without-the-tears-10bf053ea21b

### install.sh

1. **Added missing build prerequisites** (`libssl-dev autoconf automake libtool patchelf`) to both Debian and Ubuntu apt package lists. On a clean Debian install, these are not pulled in transitively and CMake's source bootstrap crashes with `Could not find OpenSSL` during `pywhispercpp` compilation.

2. **Added runtime smoke test** to `verify_installation()`. The installer now actually tries `from pywhispercpp.model import Model` before printing "Installation Complete!". If it fails, the user gets actionable diagnostics (`ldd` command, `--rebuild-whispercpp` flag, VOSK fallback) instead of a deceptive success message.

3. **Made Debian a first-class target.** The installer no longer warns "primarily designed for Ubuntu-based systems" on Debian. Debian-specific package lists and Ayatana AppIndicator handling were already in place — the messaging just hadn't caught up.

4. **Added `DEBIAN_FRONTEND=noninteractive`** to all `apt install` commands to prevent interactive debconf prompts from blocking unattended installs.

5. **Improved ydotool failure messaging** — Debian doesn't package `ydotool`, but the previous fallback was a silent "optional" message. Now the installer provides explicit compile-from-source instructions and notes that `wtype` handles most Wayland compositors.

### docs/DISTRO_COMPATIBILITY.md

6. **Added Debian-specific troubleshooting section** covering:
   - pywhispercpp build prerequisites
   - ydotool compile-from-source instructions
   - Scoped source rebuild (`--no-binary=pywhispercpp` vs `--no-binary :all:`)
   - `libwhisper.so.1` resolution diagnostics

## Why this matters

Without these fixes, a Debian user goes through:
- "Installation Complete!" → launches → `libwhisper.so.1: cannot open shared object file`
- Has to manually discover that `libssl-dev autoconf automake libtool patchelf` are missing
- Has to figure out scoped `--no-binary=pywhispercpp` vs `--no-binary :all:` to avoid the NumPy rebuild trap
- Has no clear path to getting ydotool on their system

With these fixes:
- The build deps are installed automatically
- If anything still fails, the smoke test catches it before the success banner
- The user gets specific diagnostic commands and fallback options

Fixes # (issues to be linked after creation)

### DIFF
--- a/docs/DISTRO_COMPATIBILITY.md
+++ b/docs/DISTRO_COMPATIBILITY.md
@@ -90,6 +90,53 @@ sudo apt install -y python3-gi gir1.2-gtk-3.0 gir1.2-gdkpixbuf-2.0 \
 
 **Note for Debian 11+:** Use `gir1.2-ayatanaappindicator3-0.1` instead of `gir1.2-appindicator3-0.1`
 
+### Debian-specific notes
+
+Debian's standard repositories differ from Ubuntu in a few ways that matter for Vocalinux:
+
+**pywhispercpp build prerequisites** — On a clean Debian install the following packages are not
+pulled in transitively (unlike Ubuntu) but are required when `pywhispercpp` must be compiled
+from source (e.g. for GPU support):
+
+```bash
+sudo apt install -y libssl-dev autoconf automake libtool patchelf
+```
+
+The installer now installs these automatically, but if you hit a CMake error like
+`Could not find OpenSSL` during a manual reinstall, add the above first.
+
+**ydotool** — `ydotool` is not packaged in Debian's standard repos. The installer falls back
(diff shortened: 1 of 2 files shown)


cslev:
Amazing, thanks for doing it

PR #464, opened 2026-06-10
fix(dictation): preserve spacing between silence segments

jatinkrmalik:
## Summary

- Preserve spacing between silence-separated dictation chunks in the same recording session.
- Avoid adding a leading space when the user starts a fresh dictation session.
- Store the exact text that was injected, including an automatic separator, so `delete that` remains accurate.

## What Was Happening

Vocalinux already had logic to add a space before a new text chunk when something had already been injected. The problem was that we were clearing that memory too early.

After a silence break, the recognition loop processes the completed audio chunk and then moves back to `LISTENING` while the same dictation session is still active. The previous callback cleared `last_injected_text` on every `LISTENING` state, so the next sentence looked like the first sentence again.

That caused:

```text
Hello. + World

Expected: Hello. World
Actual:   Hello.World
```

## How The Fix Works

We now clear `last_injected_text` only when recognition returns to `IDLE`, which means the dictation session has ended. During the same session, we keep the memory so later chunks get a separator.

```text
Before
------

start session
  |
  v
LISTENING
  |
  v
inject "Hello." and remember it
  |
  v
silence break
  |
  v
PROCESSING -> LISTENING
                |
                v
          clear memory here
          (too early)
                |
                v
inject "World" with no leading space

Result: Hello.World
```

```text
After
-----

start session
  |
  v
LISTENING
  |
  v
inject "Hello." and remember it
  |
  v
silence break
  |
  v
PROCESSING -> LISTENING
                |
                v
          keep memory
                |
                v
inject " World" because text already exists
  |
  v
session stops
  |
  v
IDLE -> clear memory

Result: Hello. World
Next session still starts without a leading space.
```

## Why This Keeps The First Chunk Clean

At the start of a new session, `last_injected_text` is empty. The first transcription chunk is injected as-is.

Only after a successful injection do we remember that text. Any later chunk in the same session sees that memory and gets a single leading space.

When the session ends and the engine reaches `IDLE`, the memory is cleared so the next dictation session starts cleanly again.

## Investigation

Addresses the report in https://github.com/jatinkrmalik/vocalinux/discussions/462.

The linked video is titled "VocaLinux: issue appending sentences" and has an English caption track. The transcript confirms the repro: after a pause, starting another sentence appends it directly to the previous one without whitespace.

## Tests

- `/private/tmp/vocalinux-pytest-venv/bin/python -m pytest tests/test_main.py tests/test_action_handler.py -q`

### DIFF
--- a/src/vocalinux/main.py
+++ b/src/vocalinux/main.py
@@ -409,8 +409,8 @@ def main():
         #
         #   state_callback(state: RecognitionState)
         #       Called whenever the engine transitions state (IDLE → LISTENING,
-        #       etc.).  Used here to clear the "last injected" buffer when a
-        #       new listening session starts.
+        #       etc.).  Used here to clear the "last injected" buffer after a
+        #       listening session ends.
         # ------------------------------------------------------------------
 
         def text_callback_wrapper(text: str) -> None:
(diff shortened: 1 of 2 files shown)


laurentpayot:
Thanks! :+1:

Issue #408, opened 2026-04-28
Post-transcription enhancement via cloud LLMs (OpenRouter support, custom dictionary, user-defined prompts)


avinashkanaujiya:
### Problem Statement
Vocalinux currently provides excellent offline transcription via Whisper/VOSK, but raw speech-to-text output often contains:
- Domain-specific jargon, brand names, or technical terms that local models misspell
- Grammatical artifacts from spoken language (false starts, filler words, run-on sentences)
- Lack of formatting (no punctuation standardization, capitalization fixes, or structural cleanup)

Users who need polished, publication-ready text currently have to manually edit transcriptions or pipe output through external tools, breaking the seamless "speak and type anywhere" workflow.

### Proposed Solution
Introduce an optional **post-transcription enhancement layer** powered by cloud LLMs, with the following capabilities:

1. **Cloud LLM Integration**
   - Send raw transcription text to a configurable cloud LLM API for refinement.
   - Support generic OpenAI-compatible endpoints so users can bring their own provider.

2. **First-Class OpenRouter Support**
   - Native support for [OpenRouter](https://openrouter.ai/) as a provider, allowing users to access hundreds of models (Claude, GPT, Llama, Mistral, etc.) with a single API key.
   - Expose model selection (e.g., `anthropic/claude-sonnet-4-20250514`, `openai/gpt-4.1-mini`) in settings.

3. **Custom User Prompts**
   - Allow users to define a system and/or user prompt template that receives the raw transcription.
   - Include a placeholder (e.g., `{{transcription}}`) that gets substituted at runtime.
   - Provide sensible defaults (e.g., "Fix grammar and punctuation, preserve meaning") while allowing full customization.

4. **Dictionary / Vocabulary Support**
   - Maintain a per-user custom dictionary (word/phrase → preferred spelling or expansion).
   - The dictionary should be applied *before* or *alongside* the LLM pass so that:
     - Proper nouns, technical terms, and acronyms the user cares about are preserved correctly.
     - The LLM prompt can optionally reference the dictionary to enforce consistency.
   - Format: simple JSON or TOML file, editable via UI or config file.

5. **Workflow Integration**
   - Make the enhancement step **opt-in** and **async** so it doesn't block real-time typing.
   - Indicator in the UI when enhancement is pending or complete.
   - Respect user privacy: clear warnings that cloud LLMs send data off-device; default to offline-only unless explicitly enabled.

### Use Case
[...]

PR #479, opened 2026-06-24
Pipe transcriptions through an optional postprocessing script

karottenreibe:
## Description

Add a configurable post-processing step that passes each transcription
    result to an external executable via stdin and uses its stdout as the
    text to type. Enables custom transforms (formatting, LLM rewriting, etc.)
    without modifying Vocalinux. Falls back to original text on script
    failure or timeout.

## Related Issue

None

## Type of Change

- [ ] 🐛 Bug fix (non-breaking change which fixes an issue)
- [x] ✨ New feature (non-breaking change which adds functionality)
- [ ] 📖 Documentation update
- [ ] 🧹 Code refactoring
- [ ] ✅ Test update

## Checklist

- [x] Pre-commit hooks pass

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -281,9 +281,14 @@ Configuration is stored in `~/.config/vocalinux/config.json`:
     "model_size": "tiny",
     "vad_sensitivity": 3,
     "silence_timeout": 2.0
+  },
+  "post_processing": {
+    "script_path": ""
   }
 }
 ```
+
+Set `post_processing.script_path` to the path of an executable to transform each transcription result (stdin → stdout) before it is injected. Leave empty to disable.
 
 For whisper.cpp, `model_size` may be a size such as `tiny` or an exact ggml model ID
 such as `medium.en-q5_0` or `large-v3-turbo`. You can also configure this through
--- a/docs/USER_GUIDE.md
+++ b/docs/USER_GUIDE.md
@@ -158,6 +158,28 @@ To check which backend is being used, look for these log messages when starting 
 [INFO] whisper.cpp configured with n_threads=16
 ```
 
+### Post-Processing
+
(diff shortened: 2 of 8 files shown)


chsa-admin:
Would be great if you could assign a separate hotkey for this so only used when needed!  (just a thought - maybe not necessary if fast enough and works well) - I would add as phase II maybe

jatinkrmalik:
@karottenreibe Thanks for the PR. Piping transcriptions through an external script is a clean hook, and I like the shape of it.

A couple of things before this lands, though.

First, on where the setting lives. I've been planning to refactor the settings dialog for a while. It's gotten cramped, and injection-related options don't really have a proper home. I want to rework it into something more intuitive, with a dedicated sub-panel for text-injection and post-processing settings. So I'm hesitant to merge this as-is and slot the new toggle wherever it fits today, only to move it again shortly after. I'd rather hold onto this PR until that refactor goes through, so the setting lands in the right place the first time.

Second, on examples. Right now this ships as a building block, and that's fine, but I think it would shine more if it came with a small set of example scripts the user could pick from. Even just a couple of canonical ones (strip filler words, fix capitalization) would make the feature discoverable and make the contract between Vocalinux and the script obvious.

There's also the local-LLM angle, which actually ties into #408. The ask there is basically a quick LLM pass-through so mid-corrections and rough speech come out clean on the other end, the kind of thing WisprFlow does in the cloud but done locally. Wrapping a tiny Ollama-hosted model behind this post-processing hook feels like a natural fit and could land right in that ballpark. I'd love to see the example scripts explore that direction too.

So for now, keeping this PR open and on hold until the settings refactor goes through. Not closing it. I am interested, I just want to give the setting a proper home and a bit more meat before it goes in. The tests and docs are a nice touch.

Issue #478, opened 2026-06-23
[BUG] IBus backend false-positive on niri/sway Wayland so text injection silently no-ops


subsy:
### Summary

On Wayland compositors that don't wire applications to IBus (niri, sway, and other wlroots compositors without GTK_IM_MODULE/QT_IM_MODULE set), Vocalinux selects the IBus text-injection backend, calls commit_text(), gets OK back, and logs "Text injection completed successfully" — but no text appears in any application. Transcription works perfectly; only injection silently fails.

### Root cause

is_ibus_active_input_method() (text_injection/ibus_engine.py) has a fallback for the case where no IM env vars are set (intended for KDE Plasma): if ibus-daemon is running and get_current_engine() returns any non-empty engine, it assumes IBus is the active input method.

On these compositors ibus-daemon is often running with only a keyboard-layout engine (e.g. xkb:gb::eng) while no application is an IBus client. The heuristic returns a false positive, so TextInjector chooses WAYLAND_IBUS. IBus.commit_text() succeeds at the IBus layer regardless of whether any client consumes it, so inject_text() returns True and the failure is invisible.

### Environment

  - Compositor: niri (Wayland)
  - XDG_SESSION_TYPE=wayland, GTK_IM_MODULE/QT_IM_MODULE/XMODIFIERS all empty
  - ibus-daemon running, ibus engine → xkb:gb::eng
  - ydotool + ydotoold + /dev/uinput available and working

### Logs (abridged)

  Using ydotool for Wayland text injection
  Starting IBus warmup in background
  IBus engine process started successfully
  Using IBus for wayland-ibus text injection (best compatibility)
  ...
  whisper.cpp transcribed: 'Testing one two three'
  Starting IBus text injection: 'Testing one two thre...' (length: 21)
  Text injection completed successfully      <-- nothing actually inserted

### Suggested fix

A bare xkb:* engine is only a keyboard-layout passthrough, not a real input method that delivers commit_text() to apps. Require a real IM engine before treating IBus as active:

  if is_ibus_daemon_running():
      engine = get_current_engine()
      if engine and not engine.startswith("xkb:"):
          return True

This keeps the KDE Plasma case working (real engines like anthy, libpinyin, hangul still match) while letting niri/sway fall back to ydotool/wtype, which work. Verified locally: with the guard, is_ibus_active_input_method() returns False and injection works via ydotool.

### Possibly related

[...]

Issue #474, opened 2026-06-13
[BUG] 0.12.0-beta changes keyboard layout from german to US


bcs-23:
I am on Ubuntu 24.04. with german keyboard layout. Before dictation:

```
setxkbmap -query
rules:      evdev
model:      pc105+inet
layout:     de
```

After dictation this changes to:
```
setxkbmap -query
rules:      evdev
model:      pc105+inet
layout:     us
```

But localectl status never changes. It stays on german, even after dictation.
```
localectl status
System Locale: LANG=de_DE.UTF-8
   VC Keymap: (unset)         
   X11 Layout: de
   X11 Model: pc105
```

karottenreibe:
I managed to fix this on my system as follows:

Run `ibus-setup`, Input Method tab, make sure your input method is listed and at the top.
Then run: `ibus-daemon restart`

This made ibus stay in my preferred layout.

It would be nice if either vocalinux detected this and warned or if the install guides mentioned these steps. I think by default, ibus-setup just lists no input methods on most non-EN-US systems.

bcs-23:
@karottenreibe Thanks a lot. I tried your recommendation, for me it was only the first half of the puzzle. I also had to check "use system keyboard layout" under the advanced tab. Now everything is working correctly.

fsioni:
Reproduced this and tracked down the mechanism — it explains why `setxkbmap -query` flips to `us` while `localectl` stays on `de`.

It's the **IBus injection path**. When Vocalinux injects via IBus it briefly switches to its own engine and then calls `restore_xkb_layout()` (in `text_injection/ibus_engine.py`), which shells out to `setxkbmap`. On Wayland, `setxkbmap` doesn't touch the compositor/`localectl` state — it rewrites the **XWayland** X11 layout. Switching to the engine resets that layout to its default (`us`), and the restore either captures the wrong layout or doesn't fully re-apply it, so XWayland is left on `us`.

Consequences:
- `localectl` / GNOME native Wayland apps stay correct (`de`/AZERTY/etc.).
- **XWayland (X11) apps** — many Electron apps, older programs — inherit the `us` layout, so physical typing there becomes QWERTY. That's the "some apps changed, others didn't" symptom.

Workarounds:
- Resync the XWayland layout for the current session: `setxkbmap de` (or `fr -variant azerty`, etc.).
[...]

Issue #380, opened 2026-04-03
[Wayland] IBus text injection reports success but no text appears on KDE Plasma (Garuda/Arch)


eiseleb47:
## 🐛 Bug Description
On KDE Plasma Wayland, there are two related issues with IBus-based text injection:

IBus not detected without legacy environment variables: When IBus is correctly configured via KDE's Virtual Keyboard setting (the recommended Wayland approach), Vocalinux does not detect it as the active input method and falls back through the fallback chain (wtype → ydotool → xdotool/XWayland). The detection appears to check for GTK_IM_MODULE and QT_IM_MODULE environment variables, but on Wayland IBus itself explicitly warns users to unset these variables.
IBus engine activates but text never appears: When working around the detection issue by manually setting the legacy env vars (GTK_IM_MODULE=ibus, QT_IM_MODULE=ibus, XMODIFIERS=@im=ibus) and running ibus-daemon -drx, Vocalinux detects IBus, registers its engine, and reports "Text injection completed successfully" — but no text actually appears in the target application. Speech recognition works correctly (whisper.cpp transcribes fine). The XWayland/xdotool fallback path does inject text successfully.

Additionally, the log shows that Vocalinux saves the current engine as xkb:de::deu (German) but then restores us — this keyboard layout mismatch may be related.

## 📋 Steps to Reproduce

Install KDE Plasma on an Arch-based distro (Garuda Linux in my case)
Configure IBus as the Virtual Keyboard in KDE System Settings → Input Devices → Virtual Keyboard → IBus Wayland
Ensure GTK_IM_MODULE and QT_IM_MODULE are not set (as IBus recommends on Wayland)
Launch Vocalinux
Observe that IBus is not detected and Vocalinux falls through to XWayland fallback (text injection works via this fallback)
Now set GTK_IM_MODULE=ibus, QT_IM_MODULE=ibus, XMODIFIERS=@im=ibus and run ibus-daemon -drx
Launch Vocalinux again
Double-tap Ctrl to start recording, speak, double-tap Ctrl to stop
Observe that text is transcribed but does not appear in the focused application despite "Text injection completed successfully" in the log

## ✅ Expected Behavior

Vocalinux should detect IBus as active when configured via KDE's Virtual Keyboard setting on Wayland, without requiring legacy environment variables
When the IBus text injection path is used, dictated text should appear in the focused application

## ❌ Actual Behavior

[...]

Issue #362, opened 2026-03-28
[BUG] The accented characters á, é, í, ó, ú, and others do not appear.


gabriel-tandil:
## 💻 Environment
- **OS**: Debian 13 Kde plasma
- **Python Version**: [e.g., 3.10]
- **Vocalinux Version**: 0.10.0.beta
- **Display Server**:  Wayland
- **Speech Engine**:  Whisper
- **Model Size**: medium

jatinkrmalik:
Hey @gabriel-tandil, thanks for the logs as they made it very easy to pinpoint exactly what's going on!

@Sbenazar is right: the transcription is spot-on, the problem is purely in how ydotool injects the text. Your system uses ydotool because IBus isn't your active input method, and ydotool works by simulating raw evdev keyboard events at the kernel level. Those only cover US ASCII (0–127). There are literally no evdev keycodes for `ú`, `é`, `ñ`, etc., so they get silently dropped, not a transcription issue at all.

As for why @laurentpayot's "crème brûlée" worked? That's likely because their setup routes through wtype or IBus instead, both of which handle Unicode properly. ydotool is the odd one out here.

**Workaround (for now):** if you install IBus and set it as your active input method, Vocalinux will use it for text injection and everything will work. You can check your current one with:
```
echo $GTK_IM_MODULE
```
If it's not `ibus`, switching to it should resolve this in the meantime.

**The actual fix:** PR #376 tries to address this at the `ydotool` level, no IBus required. 

When ydotool is active and the text contains non-ASCII characters, it now copies the text to the clipboard first and simulates Ctrl+V to paste it, completely bypassing the evdev keycode limitation. Plain ASCII text still goes through `ydotool type` directly as before, so existing setups aren't affected.

Should be in the next release!

gabriel-tandil:
Thank you @jatinkrmalik for your response. at the moment activate ibus is not possible in my situation, I will wait for a next version.
A hug, Gabriel

PR #480, opened 2026-06-25
fix(wayland): reliable clipboard copy and ydotool injection on non-US layouts

fsioni:
## Summary

Fixes two independent text-injection bugs on Wayland that, combined, made dictation produce nothing usable (no text in the focused field, empty clipboard) on a GNOME/Mutter + AZERTY setup.

### 1. Clipboard copy silently failed (`_run_clipboard_command`)

`wl-copy` (and `xclip`/`xsel`) fork a background process that keeps **owning the selection** in order to serve it. That child inherits the pipes created by `subprocess.run(..., stderr=subprocess.PIPE)`, so `run()` blocks reading stderr until the child exits — i.e. until the clipboard is *next overwritten*, which may be never. The call therefore hits `_clipboard_timeout` and `_copy_to_clipboard` returns `False`, **even though the copy actually succeeded**.

Reproduced on a real session:

```
WARNING wl-copy failed: Command '['wl-copy', '...']' timed out after 2.5 seconds
copy_to_clipboard -> False
# ...yet:
$ wl-paste
...the exact text was on the clipboard all along
```

Fix: redirect `stdout`/`stderr` to `DEVNULL` so `run()` only waits for the short-lived foreground process. Measured: `PIPE` → timeout; `DEVNULL` → returns in ~0.06 s with the clipboard correctly set. I couldn't find this one reported elsewhere.

### 2. `ydotool type` scrambles text on non-US layouts (`_inject_with_wayland_tool`)

`ydotool type` emits **positional evdev keycodes** that the compositor re-interprets through the active keyboard layout, which ydotool assumes is US QWERTY. On AZERTY/QWERTZ/Dvorak the output is garbled (e.g. `message` → `,essqge`, `avec` → `qvec`) and non-ASCII characters are dropped.

The code already used clipboard paste for *non-ASCII* text (#362); this extends it to **all** text when the tool is ydotool, since `Ctrl+V` is both layout-independent and Unicode-safe. (`ydotool key` Ctrl+V uses keycodes 29+47, which map to the same physical keys on QWERTY and AZERTY.)

This completes #164: that issue was closed but only the non-ASCII case was actually fixed (via #362) — plain ASCII still garbled on non-US layouts.

## Tests

- `tests/test_text_injector.py` and `tests/test_text_injector_ext.py` pass.
- Updated `test_ydotool_ascii_*` to reflect that ydotool now always pastes.
- Added `test_clipboard_command_does_not_capture_pipe` as a regression guard for the PIPE→DEVNULL fix.
- `black` and `isort` clean.

## Related issues

- Completes #164 (ydotool wrong characters on non-US layouts) — remaining ASCII case.
- Builds on #362 (accented characters via clipboard paste).
- Distinct from #425 (Firefox-specific clipboard fallback on the IBus path) — this is the `wl-copy`/`xclip` PIPE hang plus the `ydotool` ASCII case.
- Context: users land on the `ydotool`/clipboard path largely because the IBus backend is often a silent no-op on Wayland (#478, #380), which is what surfaced both bugs here.

## Notes / scope

- `wtype` keeps its native Unicode typing path (it speaks the Wayland text-input protocol, so it is layout- and Unicode-correct already).
- Clipboard paste necessarily overwrites the user's clipboard for ydotool dictation; this was already the behaviour for accented text and is unavoidable with ydotool on Wayland.
- Deliberately **out of scope** here: the IBus-vs-ydotool backend-selection policy (#478) and the `setxkbmap`-rewrites-XWayland-to-`us` side effect (#474) — kept separate to keep this PR to two self-contained bug fixes.

https://claude.ai/code/session_01EhyratrEE4vDWhriirUB9k

### DIFF
--- a/src/vocalinux/text_injection/text_injector.py
+++ b/src/vocalinux/text_injection/text_injector.py
@@ -331,11 +331,18 @@ class TextInjector:
         return tools
 
     def _run_clipboard_command(self, tool: str, text: str) -> bool:
+        # NB: wl-copy/xclip/xsel fork a background process that keeps owning the
+        # selection in order to serve it. That child inherits our pipes, so
+        # capturing stderr via subprocess.PIPE makes run() block until the child
+        # exits (i.e. until the clipboard is next overwritten) and then time out
+        # — even though the copy itself succeeded. Redirect to DEVNULL so run()
+        # only waits for the short-lived foreground process.
         if tool == "wl-copy":
             subprocess.run(
                 ["wl-copy", text],
                 check=True,
-                stderr=subprocess.PIPE,
+                stdout=subprocess.DEVNULL,
+                stderr=subprocess.DEVNULL,
                 text=True,
                 timeout=self._clipboard_timeout,
             )
@@ -346,7 +353,8 @@ class TextInjector:
                 ["xclip", "-selection", "clipboard"],
                 input=text,
                 check=True,
-                stderr=subprocess.PIPE,
+                stdout=subprocess.DEVNULL,
+                stderr=subprocess.DEVNULL,
                 text=True,
                 timeout=self._clipboard_timeout,
             )
@@ -357,7 +365,8 @@ class TextInjector:
                 ["xsel", "--clipboard", "--input"],
                 input=text,
                 check=True,
-                stderr=subprocess.PIPE,
+                stdout=subprocess.DEVNULL,
+                stderr=subprocess.DEVNULL,
                 text=True,
                 timeout=self._clipboard_timeout,
             )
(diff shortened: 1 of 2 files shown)


jatinkrmalik (approved):
LGTM. Went through the clipboard copy fix and the ydotool layout handling, both look right.

jatinkrmalik:
Merged, thanks a lot @fsioni! The scrambled output on AZERTY/QWERTZ layouts has tripped up a bunch of people, and that clipboard hang was a nasty one to track down. This ships in v0.13.0-beta.

PR #486, opened 2026-06-30
fix(text_injection): use wtype on compositors that don't bridge IBus to native apps

### LINKED ISSUE #485: Text injection silently does nothing on COSMIC / wlroots compositors when IBus is active
## Summary
On Wayland compositors that don't bridge IBus to native Wayland clients (COSMIC, Sway, Hyprland, Wayfire, niri, river, ...), dictation appears to do nothing: no text is injected into the focused window and no error is shown.

## Environment
- Desktop: COSMIC (Wayland) — but reproducible on any wlroots-based compositor
- Session: `XDG_SESSION_TYPE=wayland`, `XDG_CURRENT_DESKTOP=COSMIC`
- Vocalinux: v0.12.0-beta (logic also present on `main`)
- IBus active (`ibus engine` reports the `vocalinux` engine), `GTK_IM_MODULE=ibus`

## What happens
With the Vocalinux IBus engine active, `TextInjector` selects the IBus backend. `VocalinuxEngine.inject_text()` calls `self.commit_text(...)` and immediately returns `True`. On COSMIC/wlroots, native Wayland apps (e.g. `cosmic-term`) never create an IBus input context, so the commit reaches nothing — but injection is reported as successful. Dictation therefore silently no-ops in terminals and other native apps. (XWayland apps and GTK/Qt apps that load the IBus IM module still receive text, which can make it look intermittent.)

## Root cause
`_check_dependencies()` prefers IBus whenever it is the active input method, with no awareness that some compositors don't deliver IBus commits to native clients. The recent KDE Plasma Wayland handling covers the *inactive*-IBus case; the COSMIC/wlroots case (IBus active but never delivered) is not handled.

Separately, the `ydotool type ""` probe used to detect ydotool exits 0 even when `ydotoold` is not running, so a daemonless ydotool install is mis-selected instead of falling back to wtype.

## Expected
On compositors that don't bridge IBus to native apps, Vocalinux should use the virtual-keyboard path (`wtype`/`ydotool`), which works for all apps including `cosmic-term`.

## Verification
`wtype` injects correctly into `cosmic-term` on COSMIC (the compositor advertises `zwp_virtual_keyboard_manager_v1`). Forcing the wtype backend makes dictation work in all terminals.

I have a fix ready (compositor denylist + reliable `ydotoold` check + unit tests) and will open a PR referencing this issue.

bbbboother:
Fixes #485

## Problem
On Wayland compositors that don't bridge IBus to native Wayland clients (COSMIC, Sway, Hyprland, Wayfire, river, niri, ...), `VocalinuxEngine.inject_text()` calls `commit_text()` and returns `True`, but native apps such as `cosmic-term` never create an IBus input context, so the text is silently dropped. Dictation appears to do nothing, with no error.

## Fix
- Add a compositor denylist (`_wayland_compositor_bridges_ibus`) so IBus is skipped on those compositors and the virtual-keyboard path (`wtype`/`ydotool`) is used instead. A **denylist** (rather than an allowlist) keeps the existing IBus-preferred behaviour on GNOME/KDE and unrecognised desktops. This complements the recent KDE Plasma Wayland handling, which covers the mirror-image (IBus-inactive) case.
- Replace the `ydotool type ""` probe — which exits `0` even when `ydotoold` isn't running — with `_is_ydotoold_running()` (socket check, then `pgrep`), so a daemonless `ydotool` install correctly falls back to `wtype` instead of selecting a no-op `ydotool`.

## Testing
- New `TestCompositorIBusBridging` (7 tests): bridged vs unbridged desktops, X11 always bridges, the `ydotoold` check three ways, and an end-to-end case asserting that an active-IBus COSMIC session selects `wtype` and never instantiates the IBus injector.
- Full `tests/test_text_injector.py` + `tests/test_text_injector_ext.py` suite passes (120 tests). `black`/`isort` clean; `flake8` clean on changed code.

## Manual verification
Verified on a real COSMIC/Wayland session: `wtype` injects into `cosmic-term` (compositor advertises `zwp_virtual_keyboard_manager_v1`), and after the fix the running app logs `Compositor 'COSMIC' does not bridge IBus... Using wtype` and dictation works in all terminals.

### DIFF
--- a/src/vocalinux/text_injection/text_injector.py
+++ b/src/vocalinux/text_injection/text_injector.py
@@ -169,6 +169,72 @@ class TextInjector:
                 logger.warning("Could not detect desktop environment, defaulting to X11")
                 return DesktopEnvironment.X11
 
+    # Wayland compositors that do NOT bridge IBus commits to native Wayland
+    # clients. On these, an IBus engine's commit_text() reaches only XWayland and
+    # GTK/Qt apps that load the IBus IM module, while native apps (e.g.
+    # cosmic-term) receive nothing -- the injection appears to succeed but the
+    # text is silently dropped. These are the smithay/wlroots-based compositors
+    # that implement input handling without IBus text-input integration.
+    _IBUS_UNBRIDGED_COMPOSITORS = (
+        "cosmic",
+        "sway",
+        "hyprland",
+        "wayfire",
+        "river",
+        "niri",
+        "labwc",
+        "weston",
+    )
+
+    def _wayland_compositor_bridges_ibus(self) -> bool:
+        """Whether native Wayland clients receive IBus commits on this compositor.
+
+        GNOME (mutter), KDE (kwin) and most full desktops bridge IBus to native
+        Wayland clients, so an engine's commit_text() reaches the focused app. A
+        handful of compositors (see ``_IBUS_UNBRIDGED_COMPOSITORS``) do not, so on
+        those we must use a virtual-keyboard tool (wtype/ydotool) instead. We use
+        a denylist rather than an allowlist so that unrecognised desktops keep the
+        previous IBus-preferred behaviour.
+        """
+        if self.environment != DesktopEnvironment.WAYLAND:
+            # X11 / XWayland: IBus reaches apps through XIM regardless of DE.
+            return True
+        desktop = " ".join(
+            os.environ.get(var, "")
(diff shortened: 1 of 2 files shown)


jatinkrmalik (approved):
LGTM. Checked the compositor denylist and the ydotoold probe, logic looks correct.

jatinkrmalik:
Thanks @bbbboother, merged! Silent drops where injection looks like it succeeded are miserable to debug, so this is a really handy fix for COSMIC and the wlroots compositors. It'll be in v0.13.0-beta.

PR #493, opened 2026-07-01
feat(shortcuts): configurable modifier+key hotkeys (e.g. Alt+R)

subsy:
## What & why

Today the shortcut that starts/stops dictation can only be a **single modifier key** — Ctrl, Alt, or Shift (either side, or a left/right-specific variant) — used as a double-tap (toggle) or a hold (push-to-talk).

I use a **split keyboard with home-row mods**, where Ctrl/Alt/Shift sit on my home row and are pressed constantly while typing.  What I want is an ordinary combo like **Alt+R** that won't trigger during normal typing.

This PR generalizes the shortcut system to support **modifier + key combos** alongside the existing single-modifier gestures.

## What you can set now

- **Existing options, unchanged:** `ctrl+ctrl`, `alt+alt`, `shift+shift`, and the `left_`/`right_` variants.
- **New combos:** `alt+r`, `ctrl+alt+r`, `super+space`, `alt+f5`, … — one or more modifiers plus one ordinary key (letters, digits, function keys, and common named keys).
- **Both modes work with combos:** toggle = press the combo once to start, press again to stop; push-to-talk = hold it.

## How it works

- `keyboard_backends/base.py`: a real parser (`ShortcutSpec` / `parse_shortcut_spec`) replaces the fixed 9-entry allowlist, plus `is_valid_shortcut()` and a display-name generator. `parse_shortcut()` is kept for backward compatibility, and a bare modifier like `ctrl` is still rejected — so existing behaviour is unchanged.
- **evdev (Wayland)** and **pynput (X11)** backends both gained combo detection. Modifier state is tracked **globally across all keyboard devices** — that's what makes it work on a split keyboard, where the modifier and the key can be on different physical halves (separate evdev devices).
- `keyboard_shortcuts.py` validates via `is_valid_shortcut()` instead of the hard-coded list.
- **Settings dialog:** a new **Custom Shortcut** row — type a combo (`alt+r`) or click **Record** and press the keys. The preset dropdown is untouched.
- Legacy single-modifier gestures are left exactly as they were.

## Compatibility

- Existing configs keep working; the 9 preset shortcuts and their display names are unchanged.
- No config migration required.

## Tests

34 new tests (`tests/test_configurable_hotkeys.py`): parser + validation, backward-compatibility, and evdev + pynput combo detection — including push-to-talk press/release, the split-keyboard case (either Alt half satisfies the modifier), and Ctrl+letter control-character handling.

## Notes / caveats

- Like the existing shortcuts, the listener reads keys **passively** (it doesn't grab them), so the combo still reaches the focused app — pick one your apps don't use.
- Developed and tested on Wayland (niri) with the evdev backend. The GTK "Record" capture widget is the one part not covered by automated tests; the typed entry is a reliable alternative.

### DIFF
--- a/src/vocalinux/ui/keyboard_backends/evdev_backend.py
+++ b/src/vocalinux/ui/keyboard_backends/evdev_backend.py
@@ -259,9 +259,7 @@ class EvdevKeyboardBackend(KeyboardBackend):
         for modifier in spec.modifiers:
             codes = MODIFIER_KEY_CODES.get(modifier, set())
             if not codes:
-                logger.error(
-                    f"Combo shortcut '{self._shortcut}': unknown modifier '{modifier}'"
-                )
+                logger.error(f"Combo shortcut '{self._shortcut}': unknown modifier '{modifier}'")
                 return
             modifier_sets.append(set(codes))
 
@@ -271,9 +269,7 @@ class EvdevKeyboardBackend(KeyboardBackend):
 
     def _required_modifiers_held(self) -> bool:
         """True if at least one key code for every required modifier is held."""
-        return all(
-            bool(codes & self._combo_pressed) for codes in self._combo_modifier_sets
-        )
+        return all(bool(codes & self._combo_pressed) for codes in self._combo_modifier_sets)
 
     def is_available(self) -> bool:
         """Check if evdev is available and we can access a keyboard device with the modifier key."""
@@ -595,10 +591,7 @@ class EvdevKeyboardBackend(KeyboardBackend):
         if self._mode == "toggle":
             current_time = time.time()
             # Debounce so a single press can't double-fire.
-            if (
-                self.double_tap_callback is not None
-                and current_time - self.last_trigger_time > 0.5
-            ):
+            if self.double_tap_callback is not None and current_time - self.last_trigger_time > 0.5:
                 logger.debug(f"Combo {self._shortcut} toggled (evdev)")
                 self.last_trigger_time = current_time
                 threading.Thread(target=self.double_tap_callback, daemon=True).start()
--- a/src/vocalinux/ui/keyboard_backends/evdev_backend.py
+++ b/src/vocalinux/ui/keyboard_backends/evdev_backend.py
@@ -8,6 +8,7 @@ input devices, which works on both X11 and Wayland (with proper permissions).
 import errno
 import logging
 import os
+import re
 import select
 import threading
 import time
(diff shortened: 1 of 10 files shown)


src/vocalinux/ui/settings_dialog.py
```diff
@@ -1626,6 +1731,100 @@ def _build_shortcuts_section(self):
...
+        shortcut = shortcut.strip().lower()
+        if not is_valid_shortcut(shortcut):
+            self.shortcut_info_label.set_markup(
+                f"<span foreground='#e01b24'>Invalid shortcut: "
+                f"<b>{GLib.markup_escape_text(shortcut or '(empty)')}</b>. "
+                "Try a modifier + key, e.g. alt+r.</span>"
+            )
+            return
+
+        self.config_manager.set("shortcuts", "toggle_recognition", shortcut)
```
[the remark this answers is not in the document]
subsy: Good catch — this is a real bug. Fixed in 52e9065: `_on_shortcut_mode_changed` now reads the active shortcut from the saved config (`shortcuts.toggle_recognition`) instead of `self.shortcut_combo.get_active_id()`, so switching between Toggle and Push-to-Talk restarts the live listener on the actual shortcut (e.g. `alt+r`) rather than a stale preset. Config is the single source of truth here since both the preset combo and the custom-shortcut entry write to it before this runs. Added `tests/test_settings_mode_change.py` to lock the behaviour in. Thanks!

subsy:
## PR feedback processed

- ✅ **Fixed & resolved (correctness): 1** — Codex **P2**: switching Toggle↔Push-to-Talk now applies the **saved** shortcut, not the stale preset combo. `_on_shortcut_mode_changed` reads `shortcuts.toggle_recognition` from config (the single source of truth). Fixed in `52e9065`, regression test in `tests/test_settings_mode_change.py`. Thread replied + resolved.
- 🔧 **CI fix:** `Lint Python` was red (black formatting) → reformatted, now green.
- 💬 **Top-level/review-summary comments: 1** — the Codex "review" banner is boilerplate (no findings), no action needed.
- 🟦 Declined as nit / out-of-scope: 0.
- 🧪 **Verified:** flake8 (critical), black, isort, and the shortcut/settings tests pass; 36 new tests across the feature.
- 🟢 **CI:** Lint, Build, and Test Python 3.9–3.13 all green.
- 🟡 **codecov** (patch 61% / project −1.33%) is red, but it's a *coverage-delta* signal, not a failing test. The uncovered lines are (a) the **pynput** combo backend — pynput isn't importable on headless CI, so those tests skip there — and (b) the **settings-dialog GTK capture methods**, which can't execute under this repo's mocked-GTK test harness (the same reason the existing settings tests are source-string based). Everything reachable in the harness is covered (base parser, evdev combo, mode-change fix, keyname mapping).

src/vocalinux/ui/keyboard_backends/evdev_backend.py
```diff
@@ -161,6 +211,14 @@ def __init__(self, shortcut: str = DEFAULT_SHORTCUT, mode: str = DEFAULT_SHORTCU
         self.double_tap_threshold = 0.3  # seconds
         self.key_pressed_devices: set[int] = set()
 
+        # Combo (modifier+key) state. Populated by _resolve_combo_targets().
+        self._combo_main_code: Optional[int] = None
+        self._combo_modifier_sets: list[set[int]] = []
+        self._combo_all_codes: set[int] = set()
+        self._combo_pressed: set[int] = set()  # currently-held combo-relevant codes
```
jatinkrmalik, line 218: > `self._combo_pressed: set[int]`

`_combo_pressed` is not cleared in the existing `SYN_DROPPED` or device-disconnect recovery paths. If the dropped events include a modifier release, this set still says Alt is held, and the next `R` alone can toggle or start dictation. A push-to-talk session can also remain active. Could we reset or reconcile combo state in those recovery paths and add a regression test for modifier-down, dropped release, then main-key press?

src/vocalinux/ui/settings_dialog.py
```diff
@@ -1626,6 +1729,100 @@ def _build_shortcuts_section(self):
         # Update UI based on initial mode
         self._update_shortcut_ui_for_mode(current_mode)
```
jatinkrmalik, line 1731: > `self._update_shortcut_ui_for_mode(current_mode)`

Could we update the mode helper to inspect the saved shortcut? With `alt+r` selected, this initializes the label as `Double-tap the key...`, but combo toggle fires on a single press. Someone following that text will start dictation on the first press and stop it on the second. `get_shortcut_display_name(current_shortcut, current_mode)` already produces `Press Alt+R`.

jatinkrmalik (changes requested):
Hi @subsy,

Request changes. The parser and backend split are sensible, and the tests cover the important normal event paths. History check found no earlier modifier-plus-key implementation; the earlier custom-shortcut mode-switch issue was fixed in `52e9065`. I found one evdev recovery bug that can leave a combo logically held after dropped events, plus one UI mismatch for Toggle combos. Python tests and lint pass. Both Codecov checks are failing.

src/vocalinux/ui/keyboard_backends/evdev_backend.py
subsy: Fixed in 994a1fd. Added `_reset_combo_state()` (clears `_combo_pressed` and ends any live push-to-talk hold via the release callback) and call it from both recovery paths: the `SYN_REPORT`-after-`SYN_DROPPED` branch in `_monitor_devices`, and `_remove_keyboard_device` on disconnect. So a dropped/lost modifier *release* can no longer leave the combo logically held — the main key alone won't false-trigger, and a hold can't get stuck. Regression tests in `tests/test_configurable_hotkeys.py::TestEvdevComboRecovery` cover exactly modifier-down → lost-release → main-key-press (no trigger), a stuck-PTT release, and the disconnect path. Good catch, thank you.

src/vocalinux/ui/settings_dialog.py
subsy: Fixed in d7175e9. `_update_shortcut_ui_for_mode` now reads the saved shortcut from config and branches on `parse_shortcut_spec(...).is_combo`, using `get_shortcut_display_name(shortcut, mode)` — so with `alt+r` in toggle mode the label reads "Press Alt+R to start/stop voice typing" instead of the misleading "Double-tap the key…". Bare-modifier presets keep their existing "Double-tap …" wording. Tests added in `tests/test_settings_mode_change.py`. Thanks!

subsy:
Both requested changes are in, on the rebased branch (now on top of `v0.13.0-beta`):

- ✅ **evdev combo state after lost events** (`994a1fd`) — added `_reset_combo_state()` and call it from both recovery paths (`SYN_DROPPED`→`SYN_REPORT`, and device disconnect). A dropped modifier *release* can no longer leave the combo logically held, so the main key alone won't false-trigger and a push-to-talk hold can't get stuck. Regression tests in `TestEvdevComboRecovery` cover modifier-down → lost-release → main-key-press, stuck-PTT release, and disconnect. Thread resolved.
- ✅ **Toggle combo label** (`d7175e9`) — `_update_shortcut_ui_for_mode` reads the saved shortcut and uses `get_shortcut_display_name`, so `alt+r` in toggle mode shows "Press Alt+R" rather than "Double-tap the key…". Bare-modifier presets keep their existing wording. Tests added. Thread resolved.

- 🧪 **Verified:** flake8 (critical), black, isort clean; affected tests pass; full suite shows no new failures vs base (the pre-existing environmental failures are unchanged).

Thanks for the careful review. Happy to iterate further.

jatinkrmalik (approved):

PR #494, opened 2026-07-01
fix(text-injection): wait for modifiers to release before Wayland injection

subsy:
## The problem

On Wayland, injecting the transcribed text works by copying it to the clipboard and simulating **Ctrl+V** with ydotool (`ydotool key 29:1 47:1 47:0 29:0`).

If a modifier key is still physically held at the moment that Ctrl+V fires, the compositor sees a *different* chord. If **Alt** is held it becomes **Ctrl+Alt+V**, which isn't paste — so **nothing gets pasted**, even though the text is sitting on the clipboard.

This happens intermittently in normal use because the shortcut that stops dictation is itself a modifier (a toggle bound to Alt, or a combo like Alt+R). Transcription can finish in a few tens of milliseconds, so the paste often fires while the user is still holding the key from the tap that stopped recording. Fast transcriptions fail; slower ones — where the key is already released — succeed. That's the intermittency. The tell-tale symptom is: **"the text is on my clipboard but nothing pasted."**

For context, I hit this constantly using a toggle/push-to-talk shortcut on a split keyboard with home-row mods - but it isn't specific to combos. **Any** modifier-based shortcut on the ydotool clipboard-paste path can trigger it.

## The fix

Before performing a Wayland injection, briefly wait for the shortcut modifier keys to be released. It:

- polls evdev for currently-held modifier keycodes **across all keyboard devices**,
- returns **immediately** if nothing is held (the common case — no added latency),
- otherwise waits until they clear, bounded by `VOCALINUX_INJECT_MODIFIER_WAIT` seconds (default `1.0`; set `0` to disable),
- and proceeds anyway once the timeout is reached, so it can never hang.

It's best-effort: if evdev isn't available or readable, it just proceeds as before.

## Scope

Only affects the Wayland injection path (ydotool clipboard-paste, ydotool type, wtype). The IBus and X11/xdotool paths are unchanged.

## Tests

6 new tests (`tests/test_injection_modifier_wait.py`): returns immediately when nothing is held, waits until release, honours the env-var disable, gives up after the timeout (never hangs), and confirms the wait happens **before** the paste/type in `_inject_with_wayland_tool`.

### DIFF
--- a/src/vocalinux/text_injection/text_injector.py
+++ b/src/vocalinux/text_injection/text_injector.py
@@ -6,6 +6,7 @@ application, supporting both X11 and Wayland environments.
 """
 
 import logging
+import math
 import os
 import shutil
 import subprocess
@@ -990,6 +991,33 @@ class TextInjector:
             logger.debug(f"Could not read modifier key state: {e}")
         return held
 
+    _DEFAULT_INJECT_MODIFIER_WAIT = 1.0
+
+    def _injection_modifier_wait_seconds(self) -> float:
+        """Return the sanitized max seconds to wait for modifiers to release.
+
+        Reads VOCALINUX_INJECT_MODIFIER_WAIT and falls back to the default for
+        anything unusable: a non-numeric value, or a non-finite one. In
+        particular ``inf`` parses fine and is positive, so without this guard it
+        would make the wait deadline infinite and block injection forever while a
+        modifier stays held.
+        """
+        raw = os.environ.get("VOCALINUX_INJECT_MODIFIER_WAIT")
+        if raw is None:
+            return self._DEFAULT_INJECT_MODIFIER_WAIT
+        try:
+            value = float(raw)
+        except ValueError:
+            return self._DEFAULT_INJECT_MODIFIER_WAIT
+        if not math.isfinite(value):
+            logger.warning(
+                "VOCALINUX_INJECT_MODIFIER_WAIT=%r is not a finite number; " "using default %.1fs",
+                raw,
+                self._DEFAULT_INJECT_MODIFIER_WAIT,
+            )
+            return self._DEFAULT_INJECT_MODIFIER_WAIT
+        return value
+
     def _wait_for_modifiers_released(self) -> None:
         """Wait briefly for shortcut modifier keys to be released before injecting.
 
@@ -999,10 +1027,7 @@ class TextInjector:
         set to 0 to disable). Best-effort: if evdev/permissions are unavailable
         it simply proceeds.
         """
-        try:
-            max_wait = float(os.environ.get("VOCALINUX_INJECT_MODIFIER_WAIT", "1.0"))
-        except ValueError:
-            max_wait = 1.0
+        max_wait = self._injection_modifier_wait_seconds()
         if max_wait <= 0:
             return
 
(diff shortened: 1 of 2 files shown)


subsy:
Note that this PR is closely related (necessary because of) #493

subsy:
## PR feedback processed

- No inline review threads or reviewer comments were open on this PR.
- 🔧 **CI fix:** `Lint Python` was failing (black formatting on the new test file) → fixed, now green.
- 🧪 Added direct coverage for `_held_modifier_keycodes` (evdev device scanning), including the non-keyboard / unopenable-device / no-evdev / mid-read-OSError error paths. 14 tests total.
- 🟢 **CI:** Lint, Build, and Test Python 3.9–3.13 all green.
- 🟢 **codecov:** project essentially flat (−0.04%); patch coverage ~86–100%.

Nothing left open.

jatinkrmalik (dismissed):
Reviewing now..

src/vocalinux/text_injection/text_injector.py
```diff
@@ -954,6 +954,72 @@ def _inject_via_clipboard_paste(self, text: str) -> bool:
...
+        """Wait briefly for shortcut modifier keys to be released before injecting.
+
+        Returns immediately if no modifier is held (the common case), so this
+        adds no latency unless the user is still holding their toggle/PTT
+        shortcut. Bounded by VOCALINUX_INJECT_MODIFIER_WAIT seconds (default 1.0;
+        set to 0 to disable). Best-effort: if evdev/permissions are unavailable
+        it simply proceeds.
+        """
+        try:
+            max_wait = float(os.environ.get("VOCALINUX_INJECT_MODIFIER_WAIT", "1.0"))
```
jatinkrmalik: `float("inf")` gets past this check and makes `deadline` infinite, so a held modifier can block injection forever. Could we fall back to the default for non-finite values (`inf`/`nan`) and add a small regression test?

jatinkrmalik (changes requested):
Thank you for raising the PR. There is just one small change, otherwise it looks good.

src/vocalinux/text_injection/text_injector.py
subsy: Good catch — fixed in 143e9b0. `float("inf")` parses fine and is `> 0`, so it slipped past the `<= 0` guard and made `deadline` infinite, blocking injection while a modifier stayed held (`nan` was nonsensical too). Extracted `_injection_modifier_wait_seconds()`, which now falls back to the 1.0s default for any non-finite value via `math.isfinite` (and still for non-numeric input). Added regression tests in `tests/test_injection_modifier_wait.py::TestInjectionModifierWaitSeconds`, including one asserting that `inf` still terminates rather than looping forever. Thanks!

subsy:
- ✅ **Non-finite `VOCALINUX_INJECT_MODIFIER_WAIT`** (`143e9b0`) — `float("inf")` parsed cleanly and passed the `<= 0` guard, making the wait deadline infinite so a held modifier could block injection forever (`nan` was nonsensical too). Extracted `_injection_modifier_wait_seconds()`, which falls back to the 1.0s default for any non-finite value (`math.isfinite`) as well as non-numeric input. Thread resolved.
- 🧪 **Regression tests** added in `TestInjectionModifierWaitSeconds`: `inf` / `-inf` / `nan` / `Infinity` / `NaN` and non-numeric all fall back to the default, valid values pass through, and one test asserts `inf` now **terminates** rather than looping forever.
- 🧪 **Verified:** black, isort, flake8 (critical) clean; injection tests pass; no new failures vs base.

Thanks for catching this.

jatinkrmalik:
LGTM!

jatinkrmalik (approved):

PR #531, opened 2026-07-17
fix(web): make Flathub well-known file non-empty for GitHub Pages

jatinkrmalik:
## Summary
GitHub Pages returns 404 for zero-byte files. After #530 the path existed on `gh-pages` but https://vocalinux.com/.well-known/org.flathub.VerifiedApps.txt still 404'd.

Adds a minimal non-empty body so the URL is served (Flathub only needs the path present for now).

## Test plan
- [ ] After deploy: `curl -sI https://vocalinux.com/.well-known/org.flathub.VerifiedApps.txt` → 200


jatinkrmalik:
Closing — live path already serves correctly on vocalinux.com after #530 deploy; no need for the non-empty file workaround.

PR #556, opened 2026-07-20
feat(text-injection): opt-in "Preserve Clipboard" for the Wayland paste fallback 🤖🤖🤖

Nosion:
## Description

On Wayland with a non-US keyboard layout, `ydotool` cannot type text directly (it emits positional US evdev keycodes), so Vocalinux injects by copying the text to the clipboard and simulating Ctrl+V (`_inject_via_clipboard_paste`). That path unavoidably overwrites the user's clipboard, and with a clipboard manager such as KDE Klipper every dictation piles up in clipboard history. The method's own docstring acknowledged this: *"There is no attempt to restore it afterward."*

This adds an **opt-in** `text_injection.restore_clipboard_after_paste` option and a **"Preserve Clipboard"** toggle in **Settings → General** (next to "Copy to Clipboard"). When enabled, the clipboard-paste path:

- marks the transient dictation copy with `wl-copy --sensitive` — **feature-probed** from `wl-copy --help` and cached, so an older `wl-clipboard` (< 2.2) without the flag silently falls back to a plain copy — so clipboard managers like Klipper keep it out of history; and
- reads the user's previous clipboard beforehand and **restores it** after the paste.

It is a no-op when `copy_to_clipboard` is enabled (there the user wants the text kept), and has **no effect on X11**, which types via `xdotool` and never reaches this path.

## Related Issue

No open issue; addresses the limitation documented in `_inject_via_clipboard_paste`'s docstring.

## Type of Change

- [x] ✨ New feature (non-breaking change which adds functionality)

## Checklist

- [x] My code follows the code style of this project (black, isort, flake8 clean)
- [ ] I have updated the documentation accordingly — the sibling "Copy to Clipboard" toggle is documented only via its in-UI tooltip; this toggle follows the same pattern with a detailed tooltip. Happy to add a USER_GUIDE section if preferred.
- [x] All new and existing tests pass locally (`tests/test_text_injector.py`; pre-existing unrelated failures in the suite are environmental)
- [ ] Pre-commit hooks — ran `black`/`isort`/`flake8` manually instead (all clean)

## Compatibility notes

- **Default off** → purely additive, no behavior change for existing users.
- **X11-safe by construction** (the X11/type path never calls the clipboard-paste method).
- **Old-KDE / old-wl-clipboard safe**: `--sensitive` is version-probed; clipboard managers that ignore the hint, or systems missing a clipboard *reader*, degrade gracefully to today's behavior (restore is skipped rather than wiping content it never captured).
- Restore also works with `xclip`/`xsel`, not only `wl-paste`.

## Testing

Tested live on **KDE Plasma Wayland** (fcitx5, Danish layout, ydotool paste):

- Toggle **on** → dictations no longer appear in Klipper history and the previous clipboard is restored after paste.
- Toggle **off** → original behavior returns.
- Debug log confirms the two code paths (`user clipboard will be saved and restored` vs `temporarily overwritten`).

New unit tests cover: default (no `--sensitive`, no read/restore), sensitive+restore, older-`wl-clipboard` fallback, `copy_to_clipboard` interaction, the `--sensitive` probe, and the empty-clipboard read.


jatinkrmalik (changes requested):
### What I checked
- Default is off, so existing behavior stays put
- `--sensitive` is feature-probed and cached; older wl-clipboard still restores
- Preserve is skipped when Copy to Clipboard is on
- `finally` restores even if the paste key chord fails
- Unit coverage is thorough; full `test_text_injector.py` passed locally

### Nits
1. `wl-paste --no-newline` will drop a trailing newline from the saved clipboard. If you can, read without `--no-newline` so restore is byte-faithful for text.
2. The 150ms sleep is best-effort. Fine for v1 if the tooltip/docs say so.
3. `codecov/patch` is red on the PR, please fix.

Nosion:
Thanks for the review! Addressed in 9884b6f:

- **Coverage:** added unit tests for the new clipboard helpers — patch coverage of the feature is now 100% (`_should_restore_clipboard`, `_wl_copy_supports_sensitive`, `_read_clipboard`, `_restore_clipboard`).
- **Best-effort timing:** documented in the `_restore_clipboard` docstring and the settings tooltip.

On **`wl-paste --no-newline`** — I think `--no-newline` is actually the byte-faithful option here. Plain `wl-paste` *appends* a trailing newline (and doubles an existing one), whereas `--no-newline` returns the selection exactly. Quick check on wl-clipboard 2.3.0:

```
clipboard "ABC"   (no newline):  wl-paste → "ABC\n"   | --no-newline → "ABC"
clipboard "XYZ\n" (has newline): wl-paste → "XYZ\n\n" | --no-newline → "XYZ\n"
```

So `--no-newline` preserves the clipboard exactly on restore; dropping it would add a spurious newline. I've kept it and added a comment explaining why — happy to revisit if you're seeing different behavior on another wl-clipboard version.

jatinkrmalik:
@Nosion CI is red. Please fix.

jatinkrmalik:
@Nosion We have some merge conflicts here. Can you please fix them as well?

Nosion:
Synced with the latest `main` again — CI is green and patch coverage is 100%.

All three review points are addressed:
- **Coverage:** added unit tests for the new clipboard helpers (was the red `codecov/patch`).
- **Best-effort timing:** documented in the `_restore_clipboard` docstring and the settings tooltip.
- **`wl-paste --no-newline`:** kept, with a comment explaining it's the byte-faithful read (plain `wl-paste` appends/doubles a trailing newline) — see the note above.

The recent settings refactor moved "Copy to Clipboard" into the Recognition → Output group, so "Preserve Clipboard" now lives right beside it. Ready for another look whenever you have a moment — thanks! 🙏

jatinkrmalik:
Hey @Nosion, I'm going to close this rather than ask for another rebase.

Clipboard restore after the ydotool paste fallback is already on `main`, and it's stricter than this branch. #588 restores after paste by default (no Settings toggle). #646 adds overlapping-paste generation counters, text-only reads, and skip-if-the-user-copied-something-else.

This PR still adds `restore_clipboard_after_paste: false` and a Preserve Clipboard switch, so merging it on current `main` would make restore opt-in again. GitHub is also still conflicting on `text_injector.py` / `config_manager.py` / `settings_dialog.py`, which are the files those two commits rewrote.

The one piece `main` does not have is `wl-copy --sensitive`, so Klipper doesn't keep the transient dictation copy. If that's still annoying, a small PR against current `main` with just that flag would be welcome (gated so Copy to Clipboard stays a normal, non-sensitive offer). No toggle, no second restore path.

Thanks for the KDE testing and for chasing the earlier nits (`--no-newline` is the right `wl-paste` read).

PR #560, opened 2026-07-21
fix: correct ydotool service setup guidance 🤖🤖🤖

### LINKED ISSUE #557: [BUG] ydotoold runs (and intermittently crash-loops) in the GDM login screen
**Date observed:** 2026-07-20
**System:** Ubuntu 26.04, kernel 7.0.0-28-generic, GNOME 50.1 (Wayland), GDM3
**Vocalinux setup:** installed in `~/.local/share/vocalinux` (venv), ydotool 1.0.4-3 (Ubuntu package) as Wayland injection fallback
**Vocalinux version:** 0.14.0b0 installed; defect verified still present in
**v0.14.2** (latest release, 2026-07-18) — this is not fixed by upgrading.

## Summary

Following the commonly-recommended setup for ydotool as Vocalinux's Wayland
text-injection backend, `ydotool.service` ends up **globally enabled for every
user session** — including the GDM greeter (`gdm-greeter`, uid 60578). The
greeter's instance has no reliable permission to `/dev/uinput`, so on some
boots it fails with `Permission denied` and systemd (`Restart=always`)
restart-loops it. Even when it starts successfully, an input-injection daemon
with a virtual keyboard device is left running inside the login screen, where
nothing needs it.

## Symptom (journal, boot of 2026-07-20 07:42)

```
07:42:10 systemd[2358]: Started ydotool.service        ← user session (linger) — OK
07:42:17 systemd[2968]: Started ydotool.service        ← gdm-greeter session
07:42:17 ydotoold[3003]: failed to open uinput device: Permission denied
07:42:17 systemd[2968]: ydotool.service: Main process exited, status=2
07:42:17 systemd[2968]: ydotool.service: Scheduled restart job, restart counter is at 1.
07:42:17 ydotoold[3058]: failed to open uinput device: Permission denied
07:42:17 systemd[2968]: ydotool.service: Failed with result 'exit-code'.
07:42:17 systemd[2968]: Started ydotool.service        ← 3rd attempt succeeds (see below)
07:42:17 kernel: input: ydotoold virtual device as /devices/virtual/input/input41
07:43:03 systemd[2968]: Stopped ydotool.service        ← greeter session ends after login
```

Side effect: the greeter's short-lived virtual keyboard (`input41`) also
triggers errors in input-remapper (`Request to autoload "ydotoold virtual
device" before a user told the service about their session`).

## Root cause chain

1. **Ubuntu ships ydotool as a systemd *user* unit only**
   (`/usr/lib/systemd/user/ydotool.service`). There is **no system-level
   `ydotoold.service`** in the Ubuntu package.

2. **Vocalinux's runtime error message pairs the apt install path with the
   source-build service name.** `vocalinux/text_injection/text_injector.py`
   (v0.14.x, lines 345-347) prints:

   ```
   Or install ydotool/wl-copy for fallback:
     sudo apt install ydotool
     sudo systemctl enable --now ydotoold
   ```

   These two lines are inconsistent with each other on Debian/Ubuntu:

   - `sudo apt install ydotool` installs the **distro** package, whose only
     unit is the *user-scope* `ydotool.service`.
   - `sudo systemctl enable --now ydotoold` refers to the **system-scope**
     `ydotoold.service` that upstream ydotool's *source build* installs.

   Run as printed, on a stock Ubuntu box, the second command fails with
   *"Failed to enable unit: Unit file ydotoold.service does not exist."*
   (verified on this machine: `systemctl cat ydotoold` → "No files found").

   Note this is specific to the error message. `docs/DISTRO_COMPATIBILITY.md`
   is *self-consistent* — it gives `sudo systemctl enable --now ydotoold`
   only as the last step of a from-source build sequence
   (`git clone … && sudo cmake --build … --target install`), which does
   install a system unit. The bug is that the in-app message mixes the two
   installation paths.

   **How this leads to the greeter problem (inference, not observed):** a user
   who hits the failing command reasonably reaches for a variant that does
   work, and `sudo systemctl --global enable ydotool` is the one that both
   resolves and survives reboots. That is the state this machine is in
   (symlink `/etc/systemd/user/default.target.wants/ydotool.service`, created
   2026-07-15, matching the Vocalinux setup date). The exact command typed
   was not recorded in shell history, so the causal link is plausible rather
   than proven — but the resulting misconfiguration is real and is what the
   rest of this report documents.

3. `--global` enables the unit for **every** user manager on the machine —
   including `gdm-greeter`, whose systemd instance starts the daemon each
   time the login screen comes up.

4. `gdm-greeter` is not in the `input` group, so it can only open
   `/dev/uinput` via the per-session ACL that systemd-logind applies for the
   active seat session — and **only because an unrelated package (Solaar,
   for a Logitech headset) ships a udev rule tagging uinput `uaccess`**:

   ```
   /usr/lib/udev/rules.d/60-solaar.rules:
     KERNEL=="uinput", SUBSYSTEM=="misc", TAG+="uaccess", ...
   ```

   Whether the greeter's ydotoold starts cleanly therefore depends on a
   boot-time race between service startup and logind ACL application. On
   most boots the ACL wins and the problem is invisible; on this boot the
   daemon lost twice before succeeding. On a machine *without* Solaar (or
   another package granting uinput `uaccess`), the greeter's instance would
   crash-loop for the entire time the login screen is displayed.

## Impact

- Restart churn and `Permission denied` noise in the journal on unlucky boots.
- An input-injection daemon and a ghost virtual keyboard device running in
  the login-screen context — unnecessary attack surface where no dictation
  can ever happen.
- Confuses other input tooling (input-remapper errors above).
- Works-by-accident dependence on a third-party udev rule (Solaar) for the
  greeter instance's permissions.

## Suggested fixes (for Vocalinux)

1. **Make the error message match the install path it recommends.** In
   `text_injection/text_injector.py` (~lines 345-347), the apt route should
   name the unit the apt package actually ships, in user scope:

   ```bash
   sudo apt install ydotool
   sudo usermod -aG input $USER               # then log out and back in
   systemctl --user enable --now ydotool.service
   ```

   (No `sudo` on the systemctl line; unit is `ydotool`, not `ydotoold`.)
   Keep `sudo systemctl enable --now ydotoold` only alongside the
   from-source instructions in `docs/DISTRO_COMPATIBILITY.md`, where it is
   correct. Ideally detect which unit exists rather than printing a fixed
   string, since the correct command genuinely differs by install method.

2. **Explicitly warn against `systemctl --global enable ydotool`.** Per-user
   enablement achieves the same thing for the dictation user without
   starting the daemon in the greeter or in other accounts. A one-line note
   would prevent this entire failure mode.

3. **Consider whether the systemd unit is needed at all.** Vocalinux already
   starts the daemon on demand — `text_injector.py` contains "Start ydotoold
   if needed so ydotool can inject via uinput" and logs "Started ydotoold for
   uinput text injection". If that path is reliable, the docs could drop the
   service-enablement step for the apt route entirely, and the recommendation
   becomes just the package plus `input` group membership.

4. Optionally, if Vocalinux ever ships its own unit: add `ConditionUser`
   guards so it cannot run in greeter/system accounts, and document the
   `input` group requirement, so a misconfigured instance fails once with a
   clear message instead of restart-looping.

## Proposed patch

Concrete change to `src/vocalinux/text_injection/text_injector.py` (the block
at lines 345-347 in v0.14.2). Because the correct command differs by install
method, detecting the unit is more robust than a fixed string:

```python
# Replace:
#     "Or install ydotool/wl-copy for fallback:\n"
#     "  sudo apt install ydotool\n"
#     "  sudo systemctl enable --now ydotoold\n"
#
# With something like:
    "Or install ydotool/wl-copy for fallback:\n"
    "  sudo apt install ydotool\n"
    "  sudo usermod -aG input $USER      # then log out and back in\n"
    "  systemctl --user enable --now ydotool.service\n"
    "  # (distro packages ship a *user* unit named 'ydotool'.\n"
    "  #  If you built ydotool from source, use instead:\n"
    "  #    sudo systemctl enable --now ydotoold)\n"
    "  # Do NOT use 'systemctl --global enable ydotool' — that also starts\n"
    "  # the daemon in the display-manager greeter session.\n"
```

Matching doc change: `docs/DISTRO_COMPATIBILITY.md` already gives the
system-scope command correctly for the from-source path; it just needs the
apt/user-scope variant documented alongside it, plus the `--global` warning.

## Note for already-affected machines

Fixing the message upstream prevents *new* occurrences but does **not**
remediate systems already in this state. The `--global` enablement is a
persistent symlink in `/etc/systemd/user/default.target.wants/`, owned by the
admin who created it — no Vocalinux upgrade adds, removes, or inspects it.
Affected users must undo it by hand (see below).

It would therefore be worth having Vocalinux **detect and warn** about the
condition, e.g. at startup or in a doctor/diagnostics command:

```bash
test -e /etc/systemd/user/default.target.wants/ydotool.service && \
  echo "warning: ydotool is enabled --global; it will also run in the login greeter"
```

## Related existing issues

Several reports touch the same daemon-readiness area but none appear to cover
the greeter/`--global` case: #173 ("ydotool requires the ydotoold daemon to be
running, which is not started by default"), #97 (Kubuntu, workaround given as
`sudo systemctl enable --now ydotoold`), #380 and #485 ("ydotoold not
running"), #157 (cross-distro compatibility, lists "ydotoold Daemon Not
Running" as medium-high severity; closed). Worth cross-referencing when filing,
since fixing the message text would reduce the recurring "daemon not running"
reports too.

## Local fix — APPLIED on this machine 2026-07-20

Note the journal excerpts earlier in this report were captured *before* this
fix, and document the un-remediated state (which is what the upstream issue
is about).

Post-fix state verified 2026-07-20 08:35:
`/etc/systemd/user/default.target.wants/ydotool.service` removed (nothing
referencing ydotool left anywhere under `/etc/systemd/user`); user-scope
symlink now at `~/.config/systemd/user/default.target.wants/ydotool.service`;
`systemctl --user is-enabled ydotool.service` → `enabled`; daemon healthy
(PID 2427, `NRestarts=0`, holding `/dev/uinput`, socket at
`/run/user/1000/.ydotool_socket`).

**Still to confirm on next boot** — configuration is correct, but greeter
behaviour can only be observed on a fresh boot. Check with:

```bash
journalctl -b 0 | grep -iE "ydotoold|ydotool.service"
```

Expected: ydotool started once, by your own user manager, with **no**
`failed to open uinput device: Permission denied` and no restart-counter
lines.

```bash
# stop enabling ydotoold for every user (greeter included):
sudo systemctl --global disable ydotool.service

# enable it just for the dictation user (already in the `input` group,
# and lingering is on, so it still starts at boot):
systemctl --user enable --now ydotool.service
```

Vocalinux's primary injection path on this machine is currently scoped IBus
injection (per `~/.local/state/vocalinux/vocalinux.log`), with ydotool as
fallback — the per-user daemon keeps that fallback fully working.

AnayGarodia:
## Description

Corrects the ydotool setup guidance so the Ubuntu package route uses the user-scoped unit it actually ships, while keeping the source-build system unit as a clearly separate alternative.

The Wayland dependency check now also detects the persistent `systemctl --global enable ydotool.service` symlink and emits actionable remediation before a globally enabled input-injection daemon can continue unnoticed in display-manager greeter sessions.

Documentation now explains both unit scopes, the `input` group/log-out requirement, the `--global` risk, and recovery for already affected systems.

## Related Issue

Fixes #557

## Type of Change

- [x] 📖 Documentation update
- [ ] 🧹 Code refactoring
- [x] ✅ Test update

## Checklist

- [x] All new and existing relevant tests pass locally
- [x] Pre-commit-equivalent checks pass

N/A — CLI diagnostics and documentation only.

## Additional Notes

Verification:

- `PYTHONPATH=src python -m pytest tests/test_text_injector.py tests/test_text_injector_ext.py tests/test_injection_modifier_wait.py -q` — 167 passed
- Black and isort checks pass on changed Python files
- The CI-critical flake8 selection (`E9,F63,F7,F82`) passes across `src/` and `tests/`
- `git diff --check` passes

The runtime warning uses `os.path.lexists` intentionally so a broken global-enablement symlink is still diagnosed.

### DIFF
--- a/docs/DISTRO_COMPATIBILITY.md
+++ b/docs/DISTRO_COMPATIBILITY.md
@@ -361,7 +361,7 @@ Vocalinux requires text injection tools to work with X11 or Wayland:
 ### For Wayland
 - **IBus**: Recommended for KDE Plasma Wayland and generally the most reliable direct text input path
 - **wtype**: Recommended for wlroots-style Wayland compositors such as sway
-- **ydotool**: Universal alternative that works with both X11 and Wayland, but requires `ydotoold`; Vocalinux uses clipboard paste with `wl-clipboard`, `xclip`, or `xsel` to avoid non-US layout scrambling
+- **ydotool**: Universal alternative that works with both X11 and Wayland. Current distro packages use the per-user `ydotool.service`; source builds can install the system-level `ydotoold.service`. Vocalinux uses clipboard paste with `wl-clipboard`, `xclip`, or `xsel` to avoid non-US layout scrambling.
 - **xdotool**: May work via XWayland fallback
 
 ### Installation by Distribution
@@ -372,6 +372,28 @@ sudo dnf install xdotool wtype wl-clipboard xclip xsel     # Fedora
 sudo pacman -S xdotool wtype wl-clipboard xclip xsel       # Arch
 sudo zypper install xdotool wtype wl-clipboard xclip xsel  # openSUSE
 ```
+
+Ubuntu releases which package `ydotool` install a **user** unit named
+`ydotool.service`. Enable it only for the account running Vocalinux:
+
+```bash
+sudo apt install ydotool
+sudo usermod -aG input $USER  # then log out and back in
+systemctl --user enable --now ydotool.service
+```
+
+Do **not** use `sudo systemctl --global enable ydotool.service`. Global user-unit
+enablement also starts the input-injection daemon for the display-manager greeter
+and other accounts. If it was previously enabled globally, repair the setup with:
+
+```bash
+sudo systemctl --global disable ydotool.service
+systemctl --user enable --now ydotool.service
+```
+
+The command is different for the Debian source-build path documented above: the
+upstream install supplies the system unit `ydotoold.service`, so
+`sudo systemctl enable --now ydotoold.service` is correct only for that route.
 
 ## Manual Installation
 
--- a/src/vocalinux/text_injection/text_injector.py
+++ b/src/vocalinux/text_injection/text_injector.py
@@ -26,6 +26,34 @@ from .ibus_engine import (
 
 logger = logging.getLogger(__name__)
 
+_GLOBAL_YDOTOOL_UNIT = "/etc/systemd/user/default.target.wants/ydotool.service"
+
+
+def _ydotool_install_guidance() -> str:
+    """Return package- and source-specific ydotool service instructions."""
+    return (
+        "Ubuntu package setup for ydotool:\n"
+        "  sudo apt install ydotool\n"
+        "  sudo usermod -aG input $USER  # then log out and back in\n"
+        "  systemctl --user enable --now ydotool.service\n"
+        "  Do NOT use 'systemctl --global enable ydotool.service'; it also starts "
+        "ydotool in display-manager greeter sessions.\n"
+        "If ydotool was built from source and installed a system unit, use instead:\n"
+        "  sudo systemctl enable --now ydotoold.service"
+    )
+
+
+def _warn_if_ydotool_globally_enabled() -> None:
+    """Warn when ydotool's user unit is enabled for every account, including greeters."""
+    if os.path.lexists(_GLOBAL_YDOTOOL_UNIT):
+        logger.warning(
+            "ydotool.service is enabled globally at %s. This can start an input-injection "
+            "daemon in display-manager greeter sessions. Disable it with "
+            "'sudo systemctl --global disable ydotool.service', then enable it only for "
+            "this user with 'systemctl --user enable --now ydotool.service'.",
+            _GLOBAL_YDOTOOL_UNIT,
+        )
+
 
 def _is_kde_plasma_session() -> bool:
     """Return True when the current desktop session appears to be KDE Plasma."""
@@ -404,6 +432,9 @@ class TextInjector:
             ydotool_available = shutil.which("ydotool") is not None
             xdotool_available = shutil.which("xdotool") is not None
 
+            if ydotool_available:
+                _warn_if_ydotool_globally_enabled()
+
             # Prefer ydotool when the daemon is (or can be) ready. Flatpak ships
             # ydotool for native Wayland typing; wtype needs a Wayland socket.
             if ydotool_available and self._ensure_ydotoold():
@@ -437,9 +468,7 @@ class TextInjector:
                     "\n"
                     "For KDE Plasma Wayland users: wtype is not supported. "
                     f"{_kde_wayland_ibus_hint()}\n"
-                    "Or install ydotool/wl-copy for fallback:\n"
-                    "  sudo apt install ydotool\n"
-                    "  sudo systemctl enable --now ydotoold\n"
+                    f"Or install ydotool/wl-copy for fallback:\n{_ydotool_install_guidance()}\n"
                     "Or for clipboard fallback: sudo apt install wl-copy"
                 )
                 raise RuntimeError("Missing required dependencies for text injection")
--- a/tests/test_text_injector.py
+++ b/tests/test_text_injector.py
@@ -10,7 +10,12 @@ import unittest
 from unittest.mock import MagicMock, patch
 
 # Update import path to use the new package structure
-from vocalinux.text_injection.text_injector import DesktopEnvironment, TextInjector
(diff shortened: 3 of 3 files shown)


jatinkrmalik (approved):
Thanks for working through the feedback. The split between package and source-build service names is much clearer now, and the global-enable warning is a useful addition. Looks good to merge.

PR #563, opened 2026-07-22
feat(cli): add --version option 🤖🤖🤖

### LINKED ISSUE #555: [FEATURE] --version option missing
the cli command lacks a --version option, I think it's necessary!

AnayGarodia:
## Description

Add a standard `vocalinux --version` option backed by the package's existing `__version__` value. The parser now also uses a stable program name so help and version output remain consistent across Python versions.

Closes #555

## Type of Change

- [x] ✨ New feature
- [x] ✅ Test update

## Validation

- `PYTHONPATH=src python3 -m pytest tests/test_main_args_deps.py -q` (25 passed)
- `python3 -m py_compile src/vocalinux/main.py tests/test_main_args_deps.py`

## Checklist

- [x] Code follows the existing project style
- [x] Tests cover the new behavior
- [x] Focused tests pass locally

### DIFF
--- a/src/vocalinux/main.py
+++ b/src/vocalinux/main.py
@@ -222,6 +222,10 @@ def check_appindicator_support():
 
 def main():
     """Main entry point for the application."""
+    # Parse arguments first so flags like --version work even when
+    # another instance already holds the single-instance lock
+    args = parse_arguments()
+
     # Check for single instance BEFORE any initialization
     from . import single_instance
 
@@ -248,8 +252,6 @@ def main():
 
     # Register cleanup to release lock on exit
     atexit.register(single_instance.release_lock)
-
-    args = parse_arguments()
 
     # Configure debug logging if requested
     if args.debug:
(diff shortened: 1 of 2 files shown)


jatinkrmalik (changes requested):
There is a real gap on the full CLI path: `main()` still acquires the single-instance lock **before** `parse_arguments()`. With Vocalinux already running (the common tray case), `vocalinux --version` exits 1 with "Another instance is already running" and never prints the version. I reproduced that with a held flock on `instance.lock`.

### Requested change
Please parse (or special-case `--version`) **before** `acquire_lock`, similar to how PR #568 by @webenefits handles `--toggle` / `--start` / `--stop`. A `main()`-level test with the lock held would lock this in.

Happy to approve once that ordering is fixed.

AnayGarodia:
good catch, moved parse_arguments() before the lock so --version and --help work while the app is running. added a test that runs main() with the lock held and checks --version exits 0. full suite passes

jatinkrmalik (approved):

jatinkrmalik:
Thanks for the follow-up fix. Moving arg parsing before the lock check makes sense. Looks good to merge.

PR #564, opened 2026-07-22
fix(xdotool): preserve input focus after injection 🤖🤖🤖

### LINKED ISSUE #549: xdotool path sends Escape after injection and steals focus (e.g. Telegram)
## Summary

After successful text injection via the **xdotool** path, Vocalinux sends an extra `Escape` keypress. In apps like **Telegram Desktop**, that immediately removes focus from the message input field, so you cannot continue typing without clicking back into the field.

## Environment

- Vocalinux: **0.14.2**
- Desktop: **Linux Mint / Cinnamon**
- Session: **X11** (`XDG_SESSION_TYPE=x11`)
- Layout: German (`de`)
- Injection path: xdotool (IBus was unreliable / fell back)

## Expected behavior

After dictation, the recognized text is inserted and the focused input field **keeps** keyboard focus.

## Actual behavior

Text is inserted correctly, then focus is lost (Telegram message composer unfocuses). Manual click is required to continue typing.

## Root cause

In `_inject_with_xdotool()` (`src/vocalinux/text_injection/text_injector.py`), after typing the text Vocalinux runs:

```python
# Try to reset any stuck modifiers
subprocess.run(
    ["xdotool", "key", "--clearmodifiers", "Escape"],
    ...
)
```

This was intended to clear stuck modifiers (e.g. after `ctrl+ctrl` toggle shortcuts), but **Escape is not a modifier reset** — many applications treat Escape as “cancel / leave input”, including Telegram.

The IBus injection path does **not** send Escape; only the xdotool path does. On Mint/X11 this is easy to hit when IBus is flaky and Vocalinux falls back to xdotool.

## Suggested fix

Replace the Escape keypress with releasing modifiers only, for example:

```python
subprocess.run(
    [
        "xdotool",
        "keyup",
        "--clearmodifiers",
        "Control_L",
        "Control_R",
        "Shift_L",
        "Shift_R",
        "Alt_L",
        "Alt_R",
        "Super_L",
        "Super_R",
    ],
    ...
)
```

Locally confirming: removing/replacing that Escape call fixes Telegram focus loss while still clearing stuck modifiers.

## Workaround

Patch `text_injector.py` as above and restart Vocalinux (until a release includes the fix).

AnayGarodia:
## Description

Release potentially stuck modifier keys after xdotool text injection instead of sending Escape. This prevents the post-injection cleanup from dismissing menus, dialogs, or focused fields while still avoiding modifier leakage.

Closes #549

## Type of Change

- [x] 🐛 Bug fix
- [x] ✅ Test update

## Validation

- Focused injector and argument tests: 96 passed
- Regression coverage verifies the modifier `keyup --clearmodifiers` command and ensures Escape is never sent

## Checklist

- [x] Code follows the existing project style
- [x] Tests cover the regression
- [x] Focused tests pass locally

### DIFF
--- a/src/vocalinux/text_injection/text_injector.py
+++ b/src/vocalinux/text_injection/text_injector.py
@@ -1004,10 +1004,23 @@ class TextInjector:
                         logger.error("Text injection timed out on final attempt")
                         raise
 
-            # Try to reset any stuck modifiers
+            # Release modifiers without sending Escape, which applications may
+            # interpret as a request to cancel or leave the focused input.
             try:
                 subprocess.run(
-                    ["xdotool", "key", "--clearmodifiers", "Escape"],
+                    [
+                        "xdotool",
+                        "keyup",
+                        "--clearmodifiers",
+                        "Control_L",
+                        "Control_R",
+                        "Shift_L",
+                        "Shift_R",
+                        "Alt_L",
+                        "Alt_R",
+                        "Super_L",
+                        "Super_R",
+                    ],
                     env=env,
                     stdout=subprocess.DEVNULL,
                     stderr=subprocess.DEVNULL,
--- a/tests/test_text_injector.py
+++ b/tests/test_text_injector.py
@@ -1141,6 +1141,32 @@ class TestTextInjectorEdgeCases(unittest.TestCase):
                 if env_backup is not None:
                     os.environ["DISPLAY"] = env_backup
 
+    def test_inject_with_xdotool_releases_modifiers_without_escape(self):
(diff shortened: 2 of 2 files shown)


jatinkrmalik (approved):
@AnayGarodia nice, focused fix.

Sending Escape after xdotool type is a classic way to cancel or unfocus the field you just typed into. Switching to `keyup --clearmodifiers` on the usual modifier set keeps the "unstick modifiers" intent without that side effect.

**LGTM.**

PR #568, opened 2026-07-22
feat: opt-in D-Bus activation for compositor global shortcuts

webenefits:
## What
Adds an opt-in way to start/stop dictation via an external D-Bus trigger, so a
compositor global shortcut (e.g. KDE Plasma bound to `vocalinux --toggle`) can
drive activation instead of the built-in evdev/pynput hotkey listener.

## Why
On Wayland the internal listener reads `/dev/input` (needs the `input` group and
is effectively a system-wide key reader). Delegating activation to the compositor
removes that: no `/dev/input` access and no `input` group needed just to start
dictation. Text injection (IBus) is unchanged and independent of activation.

## How
- New session-bus service `com.vocalinux.Vocalinux` (Gio/GDBus) with
  `Toggle`/`Start`/`Stop`; handlers marshal onto the GTK main thread.
- CLI `vocalinux --toggle/--start/--stop` forwards the command to a running
  instance and exits **before** the single-instance lock, so it never starts a
  second instance.
- Opt-in `shortcuts.disable_internal_hotkey` skips the internal listener.
- Settings → Shortcuts toggle for it (live-applied; greys out the built-in
  shortcut controls when enabled). Shown on all desktops — the feature is
  compositor-independent.
- `docs/USER_GUIDE.md` section and unit tests (CLI dispatch, config gate, D-Bus
  handlers, settings wiring).

**Default behaviour is unchanged** — double-tap Ctrl still works out of the box.

## Testing
- `pytest tests/test_dbus_activation.py tests/test_settings_shortcuts.py`
- Verified on KDE Plasma Wayland: with the option on, no `/dev/input` file
  descriptors are held by the process; activation works via a bound
  `vocalinux --toggle` shortcut.

### DIFF
--- a/docs/USER_GUIDE.md
+++ b/docs/USER_GUIDE.md
@@ -84,6 +84,49 @@ Vocalinux supports two shortcut modes for controlling voice typing:
 - **Push-to-talk mode**: Hold the configured shortcut key to speak, then release to stop
 - Configure mode and key in **Settings -> Shortcuts**
 
+### Activation via a KDE Plasma global shortcut
+
+Instead of the built-in hotkey listener, you can let your desktop's global
+shortcut system trigger Vocalinux. On Wayland the built-in listener reads
+`/dev/input` (requiring membership in the `input` group and effectively acting
+as a system-wide key reader). Delegating activation to the compositor avoids
+this entirely — no `/dev/input` access and no `input` group needed just to
+start/stop dictation. Text injection is unaffected.
+
+A running instance exposes a D-Bus service on the session bus
+(`com.vocalinux.Vocalinux`), and the CLI can forward commands to it:
+
+```bash
+vocalinux --toggle   # start if idle, stop if active
+vocalinux --start    # start voice typing
+vocalinux --stop     # stop voice typing
+```
+
+To use it on KDE Plasma:
+
+1. Disable the internal hotkey listener by adding this to your Vocalinux config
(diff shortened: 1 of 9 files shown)


jatinkrmalik (changes requested):
@webenefits Can you please rebase this branch to latest main? There are a bunch of merge conflicts that you need to resolve.

webenefits:
@jatinkrmalik done! Rebased onto the latest main and resolved all conflicts. The branch is now linear on top of current upstream main (through cc69c7e), with no merge commits.

The conflicts were all complementary: the external D-Bus activation feature and the newer auto_pause_monitor / model_keepalive additions touch adjacent lines in main.py, tray_indicator.py and settings_dialog.py. Both sides are kept, nothing dropped. Full test suite passes locally (Python 3.9 / 3.11 / 3.13); flake8 / black / isort clean.

Ready for another look, could you re-review when you get a chance? Thanks!

webenefits:
@jatinkrmalik rebased again onto current main today and addressed everything Greptile flagged along the way. Thanks for jumping in and fixing a few up directly too! Verified the Settings toggle end-to-end on a real KDE Plasma machine (upgraded from a working v0.14.2 install via install.sh, whisper.cpp engine): flips on, persists, and live-disables the internal listener as expected. GitHub's still showing this blocked on your July review, could you take another look / re-approve when you have a moment?

PR #569, opened 2026-07-22
fix(uninstall): remove ~/.local/bin launcher wrappers reliably

webenefits:
`uninstall.sh` only removed `~/.local/bin/vocalinux` when it was a symlink
(`[ -L ]`), but a repo/dev install writes a real file there, and
`~/.local/bin/vocalinux-gui` was never removed — both were left behind.

- Remove both wrappers via `safe_remove` in `remove_application_files`
  (handles files and symlinks).
- `safe_remove` now also handles dangling symlinks (`-e` follows links and is
  false for a dead link, so it also tests `-L`).
- `verify_uninstallation` checks both wrappers.
- Drop the redundant symlink-only block in `remove_curl_install_files`.

### DIFF
--- a/uninstall.sh
+++ b/uninstall.sh
@@ -55,8 +55,10 @@ get_vocalinux_pids() {
 safe_remove() {
     local path="$1"
     local description="$2"
-    
-    if [ -e "$path" ]; then
+
+    # -e follows symlinks (false for a broken link), so also test -L to catch
+    # dangling symlinks left behind by an install.
+    if [ -e "$path" ] || [ -L "$path" ]; then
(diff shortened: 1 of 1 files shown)


jatinkrmalik (approved):
@webenefits thanks for the careful uninstall fix.

This matches how `install.sh` actually behaves: repo and dev installs write real files under `~/.local/bin` for both `vocalinux` and `vocalinux-gui`, so the old symlink-only check left them behind. Moving both into `remove_application_files` via `safe_remove` (and teaching `safe_remove` about dangling symlinks) is the right shape.

**LGTM.**

PR #573, opened 2026-07-24
feat(packaging): add self-contained AppImage build (x86_64 + aarch64)

jatinkrmalik:
## Summary
- Adds `packaging/appimage/build.sh`, which packages Vocalinux together with everything it needs to run (Python itself, the GTK UI toolkit, the tray-icon library) into one AppImage file. Users just download it and run it, nothing to install. It builds natively for whatever CPU it's run on.
- Hooks this into the `nightly.yml` and `release.yml` workflows for both regular PCs (x86_64) and ARM machines (aarch64), so `Vocalinux-<version>-x86_64.AppImage` and `Vocalinux-<version>-aarch64.AppImage` get published alongside the existing files on every release. The ARM build runs on a free ARM machine GitHub provides and gets attached to the same release as the regular build.
- Adds a check to `unified-pipeline.yml` that builds the AppImage for both CPU types on every pull request, so this is actually proven to work by GitHub's servers rather than just assumed. (There's no way to build or test this on a Mac, so this check is the only real verification available before merge.)

This gives people another way to run Vocalinux without installing anything: download one file, make it executable, run it. No sandbox rules to satisfy and no store review, unlike the [Flathub submission](https://github.com/flathub/flathub/pull/9368). It's meant to sit alongside the existing PyPI and AUR packages, not replace them.

Note: the AppImage still needs `xdotool`/`wtype`/`ydotool` installed on the host to actually type text into other apps. That's already true for the PyPI install too (see `docs/INSTALL.md`), so it's not a new requirement.

## Test plan
- [x] The automated build check passed for the regular (x86_64) build ([see the run](https://github.com/jatinkrmalik/vocalinux/actions/runs/30058473616))
- [x] It passed for the ARM (aarch64) build too, in the same run (produced a 103MB file in about 3 minutes)
- [ ] Download both built files and actually run them on real Linux machines (x86_64 and ARM, X11 and Wayland) to confirm the tray icon, settings window, and dictation itself all work
- [ ] After this merges, confirm the nightly build actually attaches both files to the next nightly release

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -12,6 +12,9 @@ jobs:
   nightly:
     name: Build and Publish Nightly
     runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.version.outputs.version }}
+      release_tag: ${{ steps.create_release.outputs.tag }}
 
     steps:
       - name: Checkout code
@@ -86,7 +89,7 @@ jobs:
           done
           ls -la dist/
 
-      - name: Build AppImage
+      - name: Build AppImage (x86_64)
         run: |
           bash packaging/appimage/build.sh dist/*.whl "${{ steps.version.outputs.version }}" dist
           ls -la dist/*.AppImage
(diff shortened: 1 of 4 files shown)


jatinkrmalik:
Smoke-tested the AppImage locally on my machine, works great.

PR #579, opened 2026-07-24
fix(aur): depend on virtual python-pywhispercpp so the documented GPU swap works

### LINKED ISSUE #578: AUR: hard dependency on python-pywhispercpp-cpu makes the documented CUDA swap break pacman transactions
## Problem

The AUR `vocalinux` package declares a hard dependency on `python-pywhispercpp-cpu` by exact name:

```
depends=(
  ...
  'python-pywhispercpp-cpu'
  ...
)
optdepends=(
  'python-pywhispercpp-cuda: NVIDIA GPU (replace python-pywhispercpp-cpu)'
  ...
)
```

The optdepends entry tells NVIDIA users to *replace* `python-pywhispercpp-cpu` with `python-pywhispercpp-cuda`. But that replacement is impossible in a normal pacman transaction:

- `python-pywhispercpp-cuda` has `conflicts=('python-pywhispercpp' 'python-pywhispercpp-cpu' 'python-pywhispercpp-rocm')` and only `provides=('python-pywhispercpp')` — it does **not** provide the `-cpu` name.
- So removing `-cpu` to install `-cuda` breaks vocalinux's declared dependency, and pacman refuses:

```
:: python-pywhispercpp-cuda and python-pywhispercpp-cpu are in conflict. Remove python-pywhispercpp-cpu? [y/N] y
error: failed to prepare transaction (could not satisfy dependencies)
:: removing python-pywhispercpp-cpu breaks dependency 'python-pywhispercpp-cpu' required by vocalinux
```

The only escape hatch is `pacman -Rdd python-pywhispercpp-cpu` (force-removing while ignoring dependencies), which leaves the system with a permanently broken dependency record — and every subsequent `vocalinux` upgrade tries to drag `python-pywhispercpp-cpu` back in, conflicting with the installed `-cuda` package again. GPU users hit this wall on **every single update**.

## Fix

Depend on the virtual provider name instead of the concrete CPU package. All three backend packages on the AUR (`python-pywhispercpp-cpu`, `python-pywhispercpp-cuda`, `python-pywhispercpp-rocm`, all maintained by goodroot) already declare `provides=('python-pywhispercpp')`, so:

```
depends=(
  ...
  'python-pywhispercpp'
  ...
)
```

With this, any installed backend variant satisfies the dependency, the CUDA/ROCm swap works as a normal conflict-resolution transaction, and upgrades no longer fight the user's backend choice. AUR helpers (paru/yay) prompt the user to pick a provider on fresh installs, which is the standard Arch pattern for exactly this situation (cf. `pipewire-jack`/`jack2`, `libgl`, etc.).

PR with the one-line change (plus optdepends cleanup) incoming.

## Environment

- Arch Linux (CachyOS), pacman 7.x, paru
- vocalinux 0.14.2-1 (AUR), python-pywhispercpp-cuda 1.4.0-11
- Verified working: RTX 3060 Ti with `-cuda` swapped in — `whisper_backend_init_gpu: using CUDA0 backend`

Asseel-Naji:
Fixes #578

## Problem

The AUR PKGBUILD hard-depends on `python-pywhispercpp-cpu` by exact name, while its own optdepends tells NVIDIA users to *"replace python-pywhispercpp-cpu"* with `python-pywhispercpp-cuda`. That replacement is impossible in a normal pacman transaction: the `-cuda` package conflicts with `-cpu` but only provides the virtual `python-pywhispercpp` name, so removing `-cpu` breaks vocalinux's dependency and pacman aborts. The only workaround is `pacman -Rdd`, which leaves a broken dependency record that re-breaks on every vocalinux upgrade. Full details and repro output in #578.

## Fix

- `depends`: `python-pywhispercpp-cpu` → `python-pywhispercpp` (virtual name; all three AUR backend packages `-cpu`/`-cuda`/`-rocm` declare `provides=('python-pywhispercpp')`)
- `optdepends`: list all three concrete providers so users can discover the GPU variants

Any installed backend variant now satisfies the dependency, the GPU swap becomes a normal conflict-resolution transaction, and upgrades stop fighting the user's backend choice. AUR helpers (paru/yay) prompt for a provider on fresh installs — the standard Arch pattern for swappable backends (`pipewire-jack`/`jack2`, etc.).

Since `packaging/aur/vocalinux/PKGBUILD` is auto-pushed to the AUR by `release.yml` on the next `v*` tag, merging this fixes the published AUR package with no extra steps.

## Verified

On CachyOS (Arch) with an RTX 3060 Ti: with the dependency on the virtual name, installing `python-pywhispercpp-cuda` over `-cpu` resolves cleanly and vocalinux loads the CUDA backend (`whisper_backend_init_gpu: using CUDA0 backend`).

https://claude.ai/code/session_01XAF9BkcSzwcMjTzkBxdx6k

### DIFF
--- a/packaging/aur/vocalinux/PKGBUILD
+++ b/packaging/aur/vocalinux/PKGBUILD
@@ -19,7 +19,7 @@ depends=(
   'gobject-introspection'
   'python-cairo'
   'python-pynput'
-  'python-pywhispercpp-cpu'
+  'python-pywhispercpp'
   'python-pyaudio'
   'portaudio'
   'python-numpy'
@@ -42,7 +42,9 @@ makedepends=(
   'python-setuptools'
 )
 optdepends=(
-  'python-pywhispercpp-cuda: NVIDIA GPU (replace python-pywhispercpp-cpu)'
+  'python-pywhispercpp-cpu: CPU-only whisper.cpp backend (default provider)'
+  'python-pywhispercpp-cuda: NVIDIA GPU whisper.cpp backend'
+  'python-pywhispercpp-rocm: AMD GPU whisper.cpp backend'
   'python-onnxruntime: Silero VAD'
   'xdotool: X11 injection fallback'
 )


jatinkrmalik (approved):
@Asseel-Naji **LGTM.**

Issue #591, opened 2026-07-25
[FEATURE] Feature Request: Implement model offloading / keep-alive timeout to support NVIDIA Optimus (battery saving)


dapxdap:
### Description
Hi! Thank you for this amazing project. It works incredibly fast on Linux, but there is a major issue for laptop users with hybrid graphics (NVIDIA Optimus / Intel or AMD iGPU).

Currently, Vocalinux (via `pywhispercpp`) opens a persistent Vulkan/CUDA context and holds the `large`/`medium` model weights in VRAM continuously as long as the application is running in the background. Because the Vulkan context never closes, the NVIDIA GPU is forced to stay in the `active` power state and cannot drop into `D3cold` (Runtime PM sleep). This completely drains the laptop battery in 1–1.5 hours even when the user is not transcribing anything.

### Proposed Solution
It would be amazing to have a mechanism similar to what **Ollama** or **llama.cpp** uses (like a `keep_alive` timeout):
1. Load the model weights into the system RAM upon startup.
2. When the user triggers transcription (`Ctrl+Ctrl`), quickly load/activate the model in VRAM and process the audio.
3. Keep the model in VRAM for a configurable amount of time (e.g., 5 or 10 minutes of inactivity).
4. If no transcription occurs within this timeout, completely close/release the Vulkan/CUDA context and free the VRAM, allowing the dedicated GPU to enter a deep sleep state (`suspended`).

This would massively improve the experience for Linux laptop users who want to use Vocalinux on the go without sacrificing battery life.

Thank you for considering this!

jatinkrmalik:
@dapxdap Thanks for raising this! This issue has actually been on my mind for a long time. I'm going to take a crack at it now and try to get it into the next release.

I have an early build ready here if you want to take a crack at it: https://github.com/jatinkrmalik/vocalinux/issues/445

jatinkrmalik:
Hey @dapxdap, the idle model unload / keep-alive work from this issue is on main now (merged in #592).

There’s an opt-in Unload When Idle setting (under Power). After a configurable idle timeout, Vocalinux unloads the speech model so the GPU can sleep, the next hotkey reloads it and continues as usual. That should help hybrid-GPU laptops that were stuck with VRAM held open in the background.

It’s already available if you pull main. It’ll also ship in the next nightly, and in the next stable version bump.

PR #551, opened 2026-07-19
fix(vosk): add missing Italian and English-India entries to medium/large model tables

AnayGarodia:
## Bug

Fixes #550.

The reporter selected Italian (`it`) as language and the **medium** VOSK model in Settings → Speech Engine and clicked download. Instead of downloading, the app crashed. The debug log they attached shows the exact failure, happening *before* any network request:

```
2026-07-19 21:40:12,316 - vocalinux.main - ERROR - Failed to initialize Vocalinux: join() argument must be str, bytes, or os.PathLike object, not 'NoneType'
```

## Root cause

`VOSK_MODEL_INFO["medium"]["languages"]` and `VOSK_MODEL_INFO["large"]["languages"]` in `src/vocalinux/utils/vosk_model_info.py` were missing entries for `"it"` (Italian) and `"en-in"` (English-India), even though both languages *are* present in the `"small"` table, and VOSK actually publishes bigger models for them: `vosk-model-it-0.22` (1.2G) and `vosk-model-en-in-0.5` (1G), per https://alphacephei.com/vosk/models.

Trace of the crash:
1. `_init_vosk()` in `recognition_manager.py` builds `self.vosk_model_map["medium"] = VOSK_MODEL_INFO["medium"]["languages"].get("it")`. Since `"it"` isn't a key in that dict, `.get()` returns `None`.
2. `_get_vosk_model_path()` then does `model_name = self.vosk_model_map.get(self.model_size, ...)`. Since `"medium"` *is* a key in `vosk_model_map` (with value `None`), the `.get(..., default)` fallback never kicks in — `model_name` stays `None`.
3. `os.path.join(MODELS_DIR, None)` raises `TypeError: join() argument must be str, bytes, or os.PathLike object, not 'NoneType'`.
4. This propagates up through `SpeechRecognitionManager.__init__` → the settings dialog's `download_and_apply()` background thread → caught generically and shown to the user as the truncated error in the reporter's screenshot, and (on the next app launch, since the bad config had already been saved) as the fatal startup error in their debug log.

This is a data-completeness bug (missing rows in a language table), not a network/streaming/download-robustness issue — the download logic itself (`_download_vosk_model`, timeout/retry, progress reporting) is unaffected and was never reached in this failure.

[...]

PR #609, opened 2026-07-28
chore(release): prepare v0.15.0

### LINKED ISSUE #445: [FEATURE] Auto-pause when selected games or apps are running
🚀 Feature Description
Add an option to automatically pause or stop Vocalinux when selected games or applications are running.

💡 Problem Statement
When playing demanding games, Vocalinux (especially with the Large model) uses CPU and GPU resources. I want to have the full performance of my gaming PC available and don't want to manually close and reopen Vocalinux every time.

📋 Proposed Solution
Add a section in the settings where users can select apps/games (for example Overwatch). When any of these selected apps are running, Vocalinux should automatically pause/stop. When the app is closed, Vocalinux should start again automatically.

🔄 Alternatives Considered
Manually closing Vocalinux before gaming and starting it again afterwards (current workaround).

✅ Acceptance Criteria
- [ ] Users can add apps/games to a list in the settings
- [ ] Vocalinux automatically pauses when any app from the list is running
- [ ] Vocalinux automatically resumes when the app is closed

### LINKED ISSUE #549: xdotool path sends Escape after injection and steals focus (e.g. Telegram)

### LINKED ISSUE #550: [BUG] cannot download 1.5 GB vosk model
## 🐛 Bug Description
Trying to download  a 1.5 gb vosk model, an error is produced

## 📋 Steps to Reproduce
1. Go to settings
2. Click on Speech engine
3. Try to download a model > than the smaller
4.  See screenshot

## 📸 Screenshots

Image

## 💻 Environment
- **OS**: Ubuntu 24.04
- **Python Version**: 3.12
- **Vocalinux Version**: the current at the date, cannot find a command to show version
- **Display Server**: X11
- **Speech Engine**: VOSK
- **Model Size**: medium

## 📋 Debug Logs
Run `vocalinux --debug` and paste relevant log output:

after the error:

```
 vocalinux --debug
2026-07-19 21:40:11,925 - vocalinux.single_instance - INFO - Single instance lock acquired
2026-07-19 21:40:11,928 - vocalinux.main - DEBUG - Debug logging enabled
2026-07-19 21:40:12,070 - vocalinux.utils.resource_manager - DEBUG - Checking resources candidate: /home/paolo/.local/share/vocalinux/venv/lib/python3.12/site-packages/vocalinux/resources (exists: True)
2026-07-19 21:40:12,070 - vocalinux.utils.resource_manager - DEBUG - Checking resources candidate: /home/paolo/.local/share/vocalinux/venv/lib/python3.12/resources (exists: False)
2026-07-19 21:40:12,070 - vocalinux.utils.resource_manager - DEBUG - Checking resources candidate: /home/paolo/.local/share/vocalinux/venv/share/vocalinux/resources (exists: True)
2026-07-19 21:40:12,070 - vocalinux.utils.resource_manager - DEBUG - Checking resources candidate: /home/paolo/.local/share/vocalinux/resources (exists: False)
2026-07-19 21:40:12,070 - vocalinux.utils.resource_manager - DEBUG - Checking resources candidate: /usr/share/xfce4/vocalinux/resources (exists: False)
2026-07-19 21:40:12,070 - vocalinux.utils.resource_manager - DEBUG - Checking resources candidate: /usr/share/xfce/vocalinux/resources (exists: False)
2026-07-19 21:40:12,070 - vocalinux.utils.resource_manager - DEBUG - Checking resources candidate: /usr/share/gnome/vocalinux/resources (exists: False)
2026-07-19 21:40:12,070 - vocalinux.utils.resource_manager - DEBUG - Checking resources candidate: /home/paolo/.local/share/flatpak/exports/share/vocalinux/resources (exists: False)
2026-07-19 21:40:12,070 - vocalinux.utils.resource_manager - DEBUG - Checking resources candidate: /var/lib/flatpak/exports/share/vocalinux/resources (exists: False)
2026-07-19 21:40:12,070 - vocalinux.utils.resource_manager - DEBUG - Checking resources candidate: /usr/local/share/vocalinux/resources (exists: False)
2026-07-19 21:40:12,071 - vocalinux.utils.resource_manager - DEBUG - Checking resources candidate: /usr/share/vocalinux/resources (exists: False)
2026-07-19 21:40:12,071 - vocalinux.utils.resource_manager - DEBUG - Checking resources candidate: /var/lib/snapd/desktop/vocalinux/resources (exists: False)
2026-07-19 21:40:12,071 - vocalinux.utils.resource_manager - DEBUG - Checking resources candidate: /usr/share/vocalinux/resources (exists: False)
2026-07-19 21:40:12,071 - vocalinux.utils.resource_manager - DEBUG - Checking resources candidate: /usr/local/lib/vocalinux/resources (exists: False)
2026-07-19 21:40:12,071 - vocalinux.utils.resource_manager - DEBUG - Checking resources candidate: /usr/lib/vocalinux/resources (exists: False)
2026-07-19 21:40:12,071 - vocalinux.utils.resource_manager - DEBUG - Checking resources candidate: /usr/local/lib64/vocalinux/resources (exists: False)
2026-07-19 21:40:12,071 - vocalinux.utils.resource_manager - DEBUG - Checking resources candidate: /usr/lib64/vocalinux/resources (exists: False)
2026-07-19 21:40:12,071 - vocalinux.utils.resource_manager - DEBUG - Checking resources candidate: /app/share/vocalinux/resources (exists: False)
2026-07-19 21:40:12,071 - vocalinux.utils.resource_manager - DEBUG - Resource candidate score: /home/paolo/.local/share/vocalinux/venv/lib/python3.12/site-packages/vocalinux/resources (icons=4 sounds=3 total=7)
2026-07-19 21:40:12,071 - vocalinux.utils.resource_manager - DEBUG - Resource candidate score: /home/paolo/.local/share/vocalinux/venv/share/vocalinux/resources (icons=4 sounds=3 total=7)
2026-07-19 21:40:12,071 - vocalinux.utils.resource_manager - INFO - Found resources directory: /home/paolo/.local/share/vocalinux/venv/lib/python3.12/site-packages/vocalinux/resources
2026-07-19 21:40:12,181 - vocalinux.ui.logging_manager - INFO - Logging manager initialized
2026-07-19 21:40:12,181 - vocalinux.ui.logging_manager - INFO - Global logging manager initialized
2026-07-19 21:40:12,181 - vocalinux.main - INFO - Logging system initialized
2026-07-19 21:40:12,223 - vocalinux.main - DEBUG - IBus daemon started for text injection
2026-07-19 21:40:12,224 - vocalinux.ui.config_manager - INFO - Loaded configuration from /home/paolo/.config/vocalinux/config.json
2026-07-19 21:40:12,224 - vocalinux.main - INFO - Using engine=vosk (from saved config)
2026-07-19 21:40:12,224 - vocalinux.main - INFO - Using language=it (from saved config)
2026-07-19 21:40:12,224 - vocalinux.main - INFO - Using model=medium (from saved config)
2026-07-19 21:40:12,224 - vocalinux.main - INFO - Final settings: engine=vosk, language=it, model=medium
2026-07-19 21:40:12,224 - vocalinux.main - INFO - Using audio device index=6 (name=UGREEN Camera: USB Audio (hw:2,0), from saved config)
2026-07-19 21:40:12,224 - vocalinux.main - INFO - Initializing Vocalinux...
2026-07-19 21:40:12,292 - vocalinux.speech_recognition.silero_vad - INFO - Silero VAD model loaded
2026-07-19 21:40:12,292 - vocalinux.speech_recognition.recognition_manager - INFO - Using Silero neural VAD
2026-07-19 21:40:12,292 - vocalinux.speech_recognition.recognition_manager - INFO - Initializing speech recognition with vosk engine, it language and medium model
2026-07-19 21:40:12,316 - vocalinux.main - ERROR - Failed to initialize Vocalinux: join() argument must be str, bytes, or os.PathLike object, not 'NoneType'
2026-07-19 21:40:12,316 - vocalinux.main - ERROR - Please check the logs above for more details
2026-07-19 21:40:12,318 - vocalinux.single_instance - DEBUG - Single instance lock released
```

and vocalinux doesn't start

### LINKED ISSUE #553: [FEATURE] Automatically put capital letter after a period.
Now vocalinux doesn't set the capital letter after a period. It would be very useful (possibly below a setting) that the first letter after a period be a capital one.

### LINKED ISSUE #555: [FEATURE] --version option missing

### LINKED ISSUE #557: [BUG] ydotoold runs (and intermittently crash-loops) in the GDM login screen

### LINKED ISSUE #565: [FEATURE] Please add support for the Hungarian language to the program.
Please add support for the Hungarian language to the program. The Whisper model itself already supports Hungarian.

### LINKED ISSUE #566: [BUG] GtkDialog doesn't implement response — crash in first-run dialog
## 🐛 Bug Description

On first run, clicking any response button in the welcome dialog ("Yes, start on login" / "No, I'll start manually" / "Ask me later") produces a GTK error in the console. The dialog closes and the application proceeds, but the traceback is emitted.

## 📋 Steps to Reproduce

1. Remove `~/.config/vocalinux/config.json` (or install fresh)
2. Run `vocalinux`
3. The first-run welcome dialog appears
4. Click any of the three response buttons
5. The traceback appears in the terminal

## ✅ Expected Behavior

The dialog closes cleanly without errors after selecting a response.

## ❌ Actual Behavior

The dialog closes but a GTK error is emitted:

```
Traceback (most recent call last):
  File "/usr/lib/python3.14/site-packages/vocalinux/ui/first_run_dialog.py", line 112, in do_response
    Gtk.Dialog.do_response(self, response_id)
    ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
gi.repository.GLib.GError: gi-invoke-error-quark: Class GtkDialog doesn't implement response (1)
```

The offending line calls `Gtk.Dialog.do_response(self, response_id)` directly on the parent class rather than via `super()`.

## 📸 Screenshots

N/A

## 💻 Environment

- **OS**: Arch Linux
- **Python Version**: 3.14
- **Vocalinux Version**: installed from Arch `-bin` package
- **Display Server**: X11
- **Speech Engine**: whisper_cpp
- **Model Size**: tiny

## 📋 Debug Logs

Run `vocalinux --debug` and paste relevant log output:
```
2026-07-22 21:12:19,104 - vocalinux.ui.config_manager - INFO - Config file not found at /home/ravi/.config/vocalinux/config.json. Using defaults.
Traceback (most recent call last):
  File "/usr/lib/python3.14/site-packages/vocalinux/ui/first_run_dialog.py", line 112, in do_response
    Gtk.Dialog.do_response(self, response_id)
    ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
gi.repository.GLib.GError: gi-invoke-error-quark: Class GtkDialog doesn't implement response (1)
```

## 📝 Additional Context

The application continues after the error — the configuration saves and the tray icon launches normally. This traceback is cosmetic but indicates the `do_response` override in `FirstRunDialog` is not calling the parent method correctly.

### LINKED ISSUE #567: [BUG] Heap corruption (malloc) crash when selecting Bluetooth audio input device
## 🐛 Bug Description

After completing first-run setup, selecting a Bluetooth audio input device in the settings dialog crashes the application with heap corruption.

## 📋 Steps to Reproduce

1. Start `vocalinux` from a terminal
2. Complete the first-run dialog (if shown)
3. Open Settings from the tray icon
4. Select a Bluetooth audio input device from the audio device dropdown
5. The application crashes

## ✅ Expected Behavior

The Bluetooth audio device is selected and usable for voice capture.

## ❌ Actual Behavior

```
malloc(): mismatching next->prev_size (unsorted)
zsh: abort (core dumped)  vocalinux
```

Full log leading up to the crash:

```
2026-07-22 21:21:00,715 - vocalinux.main - INFO - Final settings: engine=whisper_cpp, language=auto, model=tiny
2026-07-22 21:21:00,715 - vocalinux.main - INFO - Initializing Vocalinux...
2026-07-22 21:21:00,839 - vocalinux.speech_recognition.silero_vad - INFO - Silero VAD model loaded
2026-07-22 21:21:00,839 - vocalinux.speech_recognition.recognition_manager - INFO - Using Silero neural VAD
2026-07-22 21:21:00,839 - vocalinux.speech_recognition.recognition_manager - INFO - Initializing speech recognition with whisper_cpp engine, auto language and tiny model
2026-07-22 21:21:00,878 - vocalinux.speech_recognition.recognition_manager - INFO - whisper.cpp model 'tiny' not found at /home/ravi/.local/share/vocalinux/models/whispercpp/ggml-tiny.bin. Will download when needed.
2026-07-22 21:21:00,879 - vocalinux.text_injection.text_injector - INFO - Using text injection for x11 environment
2026-07-22 21:21:00,956 - vocalinux.text_injection.text_injector - INFO - IBus is installed but not the active input method. Falling back to alternative text injection method.
2026-07-22 21:21:00,957 - vocalinux.ui.config_manager - INFO - Loaded configuration from /home/ravi/.config/vocalinux/config.json
2026-07-22 21:21:00,957 - vocalinux.ui.keyboard_backends - INFO - Detected desktop environment: x11
2026-07-22 21:21:00,957 - vocalinux.ui.keyboard_backends - INFO - Configured shortcut: ctrl+ctrl
2026-07-22 21:21:00,957 - vocalinux.ui.keyboard_backends - INFO - Configured mode: toggle
2026-07-22 21:21:00,957 - vocalinux.ui.keyboard_backends - INFO - Using pynput backend
2026-07-22 21:21:00,957 - vocalinux.ui.tray_indicator - INFO - All required resources validated successfully
2026-07-22 21:21:00,980 - vocalinux.suspend_handler - INFO - Suspend/resume handler initialized (logind D-Bus)
2026-07-22 21:21:00,981 - vocalinux.ui.tray_indicator - INFO - Setting up keyboard shortcuts with mode: toggle
2026-07-22 21:21:00,981 - vocalinux.ui.keyboard_shortcuts - INFO - Registered toggle callback for: Ctrl (either side)
2026-07-22 21:21:00,981 - vocalinux.ui.keyboard_shortcuts - INFO - Starting keyboard shortcut listener for: Ctrl (either side)
2026-07-22 21:21:00,981 - vocalinux.ui.keyboard_backends.pynput_backend - INFO - Starting pynput keyboard listener for shortcut: ctrl+ctrl (mode: toggle)
2026-07-22 21:21:00,982 - vocalinux.ui.keyboard_backends.pynput_backend - INFO - Pynput keyboard listener started successfully
2026-07-22 21:21:00,982 - vocalinux.ui.tray_indicator - INFO - Starting GTK main loop
2026-07-22 21:21:00,983 - vocalinux.ui.tray_indicator - INFO - Initializing system tray indicator
2026-07-22 21:21:00,983 - vocalinux.ui.tray_indicator - INFO - Using icon directory: /usr/lib/python3.14/site-packages/vocalinux/resources/icons/scalable
2026-07-22 21:21:00,983 - vocalinux.ui.tray_indicator - INFO - Icon directory exists: True
2026-07-22 21:21:00,983 - vocalinux.ui.tray_indicator - INFO - Available icon files: ['vocalinux-microphone-off.svg', 'vocalinux-microphone-process.svg', 'vocalinux-microphone.svg', 'vocalinux.svg']
2026-07-22 21:21:05,231 - vocalinux.ui.settings_dialog - INFO - Found 12 audio input devices
2026-07-22 21:21:05,258 - vocalinux.ui.settings_dialog - INFO - Loaded current settings: engine=whisper_cpp, language=auto, model_size=tiny, vad=3, silence=2.0
2026-07-22 21:21:05,258 - vocalinux.ui.settings_dialog - INFO - Starting dialog with settings: engine=whisper_cpp, model=tiny
2026-07-22 21:21:05,259 - vocalinux.ui.settings_dialog - INFO - Populated 2 available engines: {'vosk': False, 'whisper': False, 'whisper_cpp': True, 'remote_api': True}
2026-07-22 21:21:05,259 - vocalinux.ui.settings_dialog - INFO - Setting active engine to: Whisper_cpp
2026-07-22 21:21:05,259 - vocalinux.ui.settings_dialog - INFO - Populating model options for engine: whisper_cpp
2026-07-22 21:21:05,259 - vocalinux.ui.settings_dialog - INFO - Saved model for whisper_cpp: tiny
2026-07-22 21:21:05,313 - vocalinux.utils.whispercpp_model_info - INFO - Vulkan support detected: deviceType = PHYSICAL_DEVICE_TYPE_INTEGRATED_GPU
2026-07-22 21:21:05,316 - vocalinux.ui.settings_dialog - INFO - Populating model options for engine: whisper_cpp
2026-07-22 21:21:05,316 - vocalinux.ui.settings_dialog - INFO - Saved model for whisper_cpp: tiny
2026-07-22 21:21:10,987 - vocalinux.ui.config_manager - INFO - Saved configuration to /home/ravi/.config/vocalinux/config.json
2026-07-22 21:21:10,987 - vocalinux.speech_recognition.recognition_manager - INFO - Audio device changed from None to 14 (name: Bluetooth internal capture stream for HUAWEI FreeBuds Pro 3)
2026-07-22 21:21:10,987 - vocalinux.ui.settings_dialog - INFO - Audio device changed to: [14] Bluetooth internal capture stream for HUAWEI FreeBuds Pro 3
2026-07-22 21:21:12,063 - vocalinux.speech_recognition.recognition_manager - INFO - Using 1 channel(s) for audio test
malloc(): mismatching next->prev_size (unsorted)
zsh: abort (core dumped)  vocalinux
```

## 📸 Screenshots

N/A

## 💻 Environment

- **OS**: Arch Linux
- **Python Version**: 3.14
- **Vocalinux Version**: installed from Arch `-bin` package
- **Display Server**: X11
- **Speech Engine**: whisper_cpp
- **Model Size**: tiny

## 📋 Debug Logs

See full log under Actual Behavior above.

## 📝 Additional Context

- The crash occurs specifically when the audio device is changed to a Bluetooth device (HUAWEI FreeBuds Pro 3 in this case)
- It occurs after `Using 1 channel(s) for audio test`, which is the audio subsystem initializing capture on the new device
- Other audio devices may not trigger this crash

### LINKED ISSUE #571: [BUG] Installer is missing dependencies: gcc-12 and libgirepository-2.0-dev
## 🐛 Bug Description
The Linux installer is missing some dependencies on Ubuntu 24 LTS

## 📋 Steps to Reproduce
1. Run the installer, [as documented here](https://vocalinux.com/#install):
`curl -fsSL https://raw.githubusercontent.com/jatinkrmalik/vocalinux/main/install.sh -o /tmp/vl.sh && bash /tmp/vl.sh --interactive`
2. On Ubuntu 24 LTS it will crash because it's missing dependencies
3. sudo apt install gcc-12 libgirepository-2.0-dev

## ✅ Expected Behavior
Vocalinux installs without issue on the first attempt

## ❌ Actual Behavior
Crash during installation because of missing dependencies

## 💻 Environment
- **OS**: #28~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Wed Jul  1 15:50:57 UTC 2
- **Python Version**: 3.12.3
- **Vocalinux Version**: [v0.14.2](https://vocalinux.com/)
- **Display Server**, **Speech Engine**, **Model Size**: N/A - didn't get that far

## 📝 Additional Context
It's worth noting that Ubuntu 24 LTS already has gcc 13 installed, just not 12.  Maybe upgrade?

### LINKED ISSUE #574: [BUG] IBus text injection reports success on every dictation, but text never reaches native apps on Kubuntu/KDE Plasma Wayland (not a first-of-session race — 100% reproducible)
## Summary

On Kubuntu (KDE Plasma, Wayland), Vocalinux consistently selects the IBus text-injection backend, transcribes correctly, calls its engine's commit path, gets a success response back, and logs `Text injection completed successfully` — but no text appears in *any* focused application. This happens on **every** dictation attempt in a session (not just the first), which rules out the cold-start race fixed in #523/#533. Speech recognition itself is perfect; only injection silently no-ops.

## Environment

- Vocalinux: 0.14.2 (pip-installed via install.sh into a venv, Python 3.14)
- Distro: Ubuntu 26.04 LTS / Kubuntu (KDE Plasma), migrated in-place from a prior Ubuntu **GNOME** install (GNOME packages purged via `apt`, not a fresh OS reinstall — see "Additional context" below)
- Session: `XDG_SESSION_TYPE=wayland`, `XDG_CURRENT_DESKTOP=KDE`, `XDG_SESSION_DESKTOP=KDE`, `DESKTOP_SESSION=plasma`
- IBus: 1.5.34-rc2, single `ibus-daemon` instance confirmed (`pgrep -fa ibus-daemon`)
- IM env vars: `GTK_IM_MODULE`, `QT_IM_MODULE`, `XMODIFIERS` all **unset**
- KDE Virtual Keyboard setting: confirmed set to "IBus Wayland" (System Settings → Input Devices → Virtual Keyboard) — initially found *unset*, corrected, issue persisted
- Speech engine: whisper.cpp, tiny model, Vulkan GPU backend — transcription is fast and accurate every time
- ydotool/ydotoold available (mentioned in logs as considered, but IBus is chosen as the final "best compatibility" backend)

## Steps to Reproduce

1. Fresh Vocalinux install on Kubuntu/KDE Plasma Wayland (all prior `~/.config/vocalinux`, `~/.local/share/vocalinux` etc. deleted before this test)
2. `vocalinux --wayland --debug`
3. Focus any native text field (tested in a plain blank text file — not Chrome/Electron-specific, ruled out app-specific quirks first)
4. Dictate a sentence, toggle off
5. Repeat 2–3 times in the same running session

## Expected vs Actual

- **Expected:** dictated text appears in the focused field.
- **Actual:** nothing appears, in any of 3 consecutive attempts in the same session. Vocalinux's own log reports full success every time.

## Diagnostic steps already taken (to save round-trips)

- `pgrep -fa ibus-daemon` → exactly one instance. Ruled out "duplicate daemon" as the cause.
- Confirmed no `GTK_IM_MODULE`/`QT_IM_MODULE`/`XMODIFIERS` set.
- `ibus engine` initially returned `No global engine`. Manually set `ibus engine xkb:us::eng` as a baseline — no change in behavior.
- Found KDE's Virtual Keyboard setting was **not** set to IBus at all initially; set it correctly, restarted Vocalinux — issue persisted.
- After Vocalinux runs, `ibus engine` reports `vocalinux` (its own engine) as active and it **stays** active across multiple dictations in the session — so engine activation/selection is working correctly.
- `dbus-monitor --session "interface='org.freedesktop.IBus.InputContext'"` showed nothing — turned out to be a red herring on my part: IBus does not use the regular session bus, so this test was invalid and was abandoned in favor of Vocalinux's own `--debug` log.
- Read through `text_injection/ibus_engine.py` and `text_injection/text_injector.py` in the installed 0.14.2 package: confirmed the FocusIn-wait fix from #533 (`_ibus_ensure_client_focus`, `PING` → `FOCUSED`/`OK`/`NO_FOCUS`) is present and appears to be functioning — the socket response is `OK` (not `NO_FOCUS`/timeout) on every attempt, so commit is not being sent blind. Despite that, text still doesn't land.

## Debug log excerpt (three consecutive dictations, same session)

```
whisper.cpp transcribed: 'This is me attempting to use vocal in X again.'
Starting IBus text injection: 'This is me attemptin...' (length: 46)
Could not determine current IBus engine; activating Vocalinux engine without restore
Text injection successful
Text injection completed successfully

whisper.cpp transcribed: 'This is me attempting to use vocal index once more.'
Starting IBus text injection: ' This is me attempti...' (length: 52)
Text injection successful
Text injection completed successfully

whisper.cpp transcribed: 'And this is me attempting to use focal length for a third time.'
Starting IBus text injection: ' And this is me atte...' (length: 64)
Text injection successful
Text injection completed successfully
```

(Full `--debug` log available on request — happy to attach.)

## New finding: the running `ibus-daemon` may never be bridged to KDE's Wayland compositor at all

This is the part I think is most worth flagging. The ibus package ships two systemd `--user` units:

```
/usr/lib/systemd/user/org.freedesktop.IBus.session.GNOME.service   (enabled, WantedBy=gnome-session.target)
/usr/lib/systemd/user/org.freedesktop.IBus.session.generic.service (static)
```

There is **no KDE/Plasma-specific unit**. On this machine, `org.freedesktop.IBus.session.GNOME.service` is `enabled` but currently **inactive** (`systemctl --user status` → `Active: inactive (dead)`) — it is not what's running ibus-daemon right now.

The actual live process is:

```
ibus-daemon -x -d -r          (pid parented directly to systemd --user after reparenting)
```

`-x -d -r` is the classic invocation used by Debian/Ubuntu's `im-config` mechanism (confirmed via `im-config -m` → configured IM is `ibus`), historically designed for X11/XIM sessions, not for wiring into a Wayland compositor's native input-method protocol.

Given there's no KDE-specific systemd unit shipped by the `ibus` package, and KDE Plasma's own "IBus Wayland" virtual-keyboard setting doesn't appear to launch its own bridged daemon (it seems to just assume *some* ibus-daemon is already running and reachable), it looks plausible that the daemon this system ends up running is never actually registered as the Wayland input-method with the compositor (kwin). That would explain exactly what we're seeing: Vocalinux's own engine gets `FocusIn`/`OK` from IBus internally (satisfying the #533 fix), `commit_text()` succeeds at the IBus D-Bus layer — but the text never reaches native Wayland app surfaces because the daemon backing it was never bridged to kwin's input-method-v2/text-input-v3 protocol in the first place.

I don't have a diagnosis of *how* Vocalinux could detect "IBus is running AND bridged to the current Wayland compositor" vs. just "IBus is running" — but if there's a way (e.g., checking for a registered `zwp_input_method_v2` client, or shelling out to a KDE-specific check), that seems like it'd need to gate the `wayland_scoped_ibus` decision in `text_injector.py` (the "Bridging Wayland DEs (GNOME/KDE)... a bare xkb:* baseline is fine (#501, #504)" comment) — that assumption may not hold for Kubuntu specifically.

## Additional context

This machine was originally installed as Ubuntu with GNOME, then switched in-place to Kubuntu/KDE Plasma without a full OS reinstall (GNOME desktop packages purged via `apt`, but the base OS was not reinstalled). `gsettings-desktop-schemas` (a generic dependency, not GNOME-specific) is still present and `org.gnome.desktop.input-sources` still resolves to `[('xkb', 'us')]` — this is what the "GNOME gsettings fallback" path in `ibus_engine.py` reads, though I don't have evidence this is actively causing harm beyond confirming "bare xkb layout, no real engine," which matches what `ibus engine` already reported directly.

I mention the migration history in case it's relevant (e.g., some KDE-specific ibus wiring package that a fresh Kubuntu install would pull in but an in-place GNOME→KDE switch wouldn't), but I want to flag that I can't confirm this is GNOME-residue vs. simply how Kubuntu ships ibus integration for everyone — happy to test on a clean Kubuntu VM if that would help narrow it down.

## Workaround in use

Enabling "Copy to clipboard" in Vocalinux's Text Injection settings works as a practical workaround — since `inject_text()` reports success, the parallel clipboard-copy path fires reliably, letting me paste manually with Ctrl+V.

### LINKED ISSUE #589: [BUG]  Unable to select NVIDIA GPU with Vulkan
# whisper.cpp Vulkan backend always selects Intel iGPU instead of NVIDIA GPU

## Environment

* Vocalinux: 0.14.2
* pywhispercpp: 1.5.1.dev1+gf3b74543d
* OS: Debian 13 (Trixie)
* Desktop Environment: Cinnamon
* CPU: Intel i7-8700
* Model size: any
* Engine: whisper.cpp
* GPUs:

  * Intel UHD 630
  * NVIDIA RTX 2060

## Problem

When using the whisper.cpp engine with Vulkan enabled, Vocalinux detects both GPUs but always selects the Intel UHD 630 instead of the NVIDIA RTX 2060.

Performance is significantly slower than expected because the integrated GPU is being used.

## What I Tested

The system correctly detects both GPUs through Vulkan:

```text
ggml_vulkan: 0 = Intel(R) UHD Graphics 630 (CFL GT2)
ggml_vulkan: 1 = NVIDIA GeForce RTX 2060
```

I also verified that NVIDIA Vulkan support is working:

```bash
vulkaninfo --summary | grep deviceName
```

Output:

```text
deviceName = Intel(R) UHD Graphics 630 (CFL GT2)
deviceName = NVIDIA GeForce RTX 2060
```

I tried launching Vocalinux with:

```bash
__NV_PRIME_RENDER_OFFLOAD=1 \
__VK_LAYER_NV_optimus=NVIDIA_only \
~/.local/bin/vocalinux
```

and also attempted Vulkan device selection variables, but Vocalinux still selected the Intel GPU.

## Relevant Log Output

```text
ggml_vulkan: Found 2 Vulkan devices:
ggml_vulkan: 0 = Intel(R) UHD Graphics 630 (CFL GT2)
ggml_vulkan: 1 = NVIDIA GeForce RTX 2060

whisper_backend_init_gpu: device 0: Vulkan0 (type: 2)
whisper_backend_init_gpu: found GPU device 0: Vulkan0 (type: 2, cnt: 0)
whisper_backend_init_gpu: using Vulkan0 backend
```

## Expected Behavior

When multiple Vulkan GPUs are available, Vocalinux should either:

1. Select the more powerful discrete GPU automatically, or
2. Provide a setting to choose which Vulkan device whisper.cpp uses.

## Question

Is there currently a supported way to select the Vulkan device used by whisper.cpp from within Vocalinux?

## 📋 Debug Logs
Run `vocalinux --debug` and paste relevant log output:
```
~/.local/bin/vocalinux --debug
2026-07-24 18:54:11,285 - vocalinux.single_instance - INFO - Single instance lock acquired
2026-07-24 18:54:11,288 - vocalinux.main - DEBUG - Debug logging enabled
2026-07-24 18:54:11,409 - vocalinux.utils.resource_manager - DEBUG - Checking resources candidate: /home/user87/.local/share/vocalinux/venv/lib/python3.13/site-packages/vocalinux/resources (exists: True)
2026-07-24 18:54:11,409 - vocalinux.utils.resource_manager - DEBUG - Checking resources candidate: /home/user87/.local/share/vocalinux/venv/lib/python3.13/resources (exists: False)
2026-07-24 18:54:11,409 - vocalinux.utils.resource_manager - DEBUG - Checking resources candidate: /home/user87/.local/share/vocalinux/venv/share/vocalinux/resources (exists: True)
2026-07-24 18:54:11,409 - vocalinux.utils.resource_manager - DEBUG - Checking resources candidate: /home/user87/.local/share/vocalinux/resources (exists: False)
2026-07-24 18:54:11,409 - vocalinux.utils.resource_manager - DEBUG - Checking resources candidate: /usr/share/gnome/vocalinux/resources (exists: False)
2026-07-24 18:54:11,409 - vocalinux.utils.resource_manager - DEBUG - Checking resources candidate: /home/user87/.local/share/flatpak/exports/share/vocalinux/resources (exists: False)
2026-07-24 18:54:11,409 - vocalinux.utils.resource_manager - DEBUG - Checking resources candidate: /var/lib/flatpak/exports/share/vocalinux/resources (exists: False)
2026-07-24 18:54:11,409 - vocalinux.utils.resource_manager - DEBUG - Checking resources candidate: /usr/local/share/vocalinux/resources (exists: False)
2026-07-24 18:54:11,409 - vocalinux.utils.resource_manager - DEBUG - Checking resources candidate: /usr/share/vocalinux/resources (exists: False)
2026-07-24 18:54:11,409 - vocalinux.utils.resource_manager - DEBUG - Checking resources candidate: /usr/local/lib/vocalinux/resources (exists: False)
2026-07-24 18:54:11,409 - vocalinux.utils.resource_manager - DEBUG - Checking resources candidate: /usr/lib/vocalinux/resources (exists: False)
2026-07-24 18:54:11,409 - vocalinux.utils.resource_manager - DEBUG - Checking resources candidate: /usr/local/lib64/vocalinux/resources (exists: False)
2026-07-24 18:54:11,409 - vocalinux.utils.resource_manager - DEBUG - Checking resources candidate: /usr/lib64/vocalinux/resources (exists: False)
2026-07-24 18:54:11,409 - vocalinux.utils.resource_manager - DEBUG - Checking resources candidate: /app/share/vocalinux/resources (exists: False)
2026-07-24 18:54:11,409 - vocalinux.utils.resource_manager - DEBUG - Resource candidate score: /home/user87/.local/share/vocalinux/venv/lib/python3.13/site-packages/vocalinux/resources (icons=4 sounds=3 total=7)
2026-07-24 18:54:11,409 - vocalinux.utils.resource_manager - DEBUG - Resource candidate score: /home/user87/.local/share/vocalinux/venv/share/vocalinux/resources (icons=4 sounds=3 total=7)
2026-07-24 18:54:11,410 - vocalinux.utils.resource_manager - INFO - Found resources directory: /home/user87/.local/share/vocalinux/venv/lib/python3.13/site-packages/vocalinux/resources
2026-07-24 18:54:11,499 - vocalinux.ui.logging_manager - INFO - Logging manager initialized
2026-07-24 18:54:11,499 - vocalinux.ui.logging_manager - INFO - Global logging manager initialized
2026-07-24 18:54:11,499 - vocalinux.main - INFO - Logging system initialized
2026-07-24 18:54:11,522 - vocalinux.ui.config_manager - INFO - Loaded configuration from /home/user87/.config/vocalinux/config.json
2026-07-24 18:54:11,522 - vocalinux.main - INFO - Using engine=whisper_cpp (from saved config)
2026-07-24 18:54:11,522 - vocalinux.main - INFO - Using language=en-us (from saved config)
2026-07-24 18:54:11,522 - vocalinux.main - INFO - Using model=large-v3-turbo (from saved config)
2026-07-24 18:54:11,522 - vocalinux.main - INFO - Final settings: engine=whisper_cpp, language=en-us, model=large-v3-turbo
2026-07-24 18:54:11,523 - vocalinux.main - INFO - Initializing Vocalinux...
2026-07-24 18:54:11,596 - vocalinux.speech_recognition.silero_vad - INFO - Silero VAD model loaded
2026-07-24 18:54:11,596 - vocalinux.speech_recognition.recognition_manager - INFO - Using Silero neural VAD
2026-07-24 18:54:11,596 - vocalinux.speech_recognition.recognition_manager - INFO - Initializing speech recognition with whisper_cpp engine, en-us language and large-v3-turbo model
2026-07-24 18:54:11,734 - vocalinux.utils.whispercpp_model_info - INFO - Vulkan support detected: deviceType         = PHYSICAL_DEVICE_TYPE_INTEGRATED_GPU
2026-07-24 18:54:11,734 - vocalinux.speech_recognition.recognition_manager - INFO - whisper.cpp backend selection priority: Vulkan -> CUDA -> CPU
2026-07-24 18:54:11,734 - vocalinux.speech_recognition.recognition_manager - INFO - whisper.cpp using Vulkan GPU backend: deviceType         = PHYSICAL_DEVICE_TYPE_INTEGRATED_GPU
2026-07-24 18:54:11,742 - vocalinux.speech_recognition.recognition_manager - INFO - whisper.cpp hardware: vulkan | deviceType         = PHYSICAL_DEVICE_TYPE_INTEGRATED_GPU | RAM: 31GB
2026-07-24 18:54:11,742 - vocalinux.speech_recognition.recognition_manager - INFO - whisper.cpp model file: /home/user87/.local/share/vocalinux/models/whispercpp/ggml-large-v3-turbo.bin (1549.3 MB)
2026-07-24 18:54:11,742 - vocalinux.speech_recognition.recognition_manager - INFO - Loading whisper.cpp 'large-v3-turbo' model...
2026-07-24 18:54:11,745 - pywhispercpp.model - INFO - Initializing the model ...
whisper_init_from_file_with_params_no_state: loading model from '/home/user87/.local/share/vocalinux/models/whispercpp/ggml-large-v3-turbo.bin'
whisper_init_with_params_no_state: use gpu    = 1
whisper_init_with_params_no_state: flash attn = 1
whisper_init_with_params_no_state: gpu_device = 0
whisper_init_with_params_no_state: dtw        = 0
ggml_vulkan: Found 2 Vulkan devices:
ggml_vulkan: 0 = Intel(R) UHD Graphics 630 (CFL GT2) (Intel open-source Mesa driver) | uma: 1 | fp16: 1 | bf16: 0 | warp size: 32 | shared memory: 65536 | int dot: 0 | matrix cores: none
ggml_vulkan: 1 = NVIDIA GeForce RTX 2060 (NVIDIA) | uma: 0 | fp16: 1 | bf16: 0 | warp size: 32 | shared memory: 49152 | int dot: 1 | matrix cores: KHR_coopmat
whisper_init_with_params_no_state: devices    = 3
whisper_init_with_params_no_state: backends   = 2
whisper_model_load: loading model
whisper_model_load: n_vocab       = 51866
whisper_model_load: n_audio_ctx   = 1500
whisper_model_load: n_audio_state = 1280
whisper_model_load: n_audio_head  = 20
whisper_model_load: n_audio_layer = 32
whisper_model_load: n_text_ctx    = 448
whisper_model_load: n_text_state  = 1280
whisper_model_load: n_text_head   = 20
whisper_model_load: n_text_layer  = 4
whisper_model_load: n_mels        = 128
whisper_model_load: ftype         = 1
whisper_model_load: qntvr         = 0
whisper_model_load: type          = 5 (large v3)
whisper_model_load: adding 1609 extra tokens
whisper_model_load: n_langs       = 100
whisper_model_load:      Vulkan0 total size =  1623.92 MB
whisper_model_load: model size    = 1623.92 MB
whisper_backend_init_gpu: device 0: Vulkan0 (type: 2)
whisper_backend_init_gpu: found GPU device 0: Vulkan0 (type: 2, cnt: 0)
whisper_backend_init_gpu: using Vulkan0 backend
whisper_init_state: kv self size  =   10.49 MB
whisper_init_state: kv cross size =   31.46 MB
whisper_init_state: kv pad  size  =    7.86 MB
whisper_init_state: compute buffer (conv)   =   37.69 MB
whisper_init_state: compute buffer (encode) =   55.35 MB
whisper_init_state: compute buffer (cross)  =   39.99 MB
whisper_init_state: compute buffer (decode) =  100.04 MB
2026-07-24 18:54:13,223 - vocalinux.speech_recognition.recognition_manager - INFO - whisper.cpp configured with n_threads=3 (GPU backend: vulkan)
2026-07-24 18:54:13,223 - vocalinux.speech_recognition.recognition_manager - INFO - whisper.cpp model loaded in 1.48s (vulkan backend)
2026-07-24 18:54:13,223 - vocalinux.speech_recognition.recognition_manager - INFO - whisper.cpp engine initialized successfully.
2026-07-24 18:54:13,224 - vocalinux.text_injection.text_injector - INFO - Using text injection for x11 environment
2026-07-24 18:54:13,241 - vocalinux.text_injection.ibus_engine - DEBUG - IBus does not appear to be the active input method (GTK_IM_MODULE=not set, QT_IM_MODULE=not set, XMODIFIERS=not set, daemon running: False)
2026-07-24 18:54:13,241 - vocalinux.text_injection.text_injector - INFO - IBus is installed but not the active input method. Falling back to alternative text injection method.
2026-07-24 18:54:13,241 - vocalinux.ui.config_manager - INFO - Loaded configuration from /home/user87/.config/vocalinux/config.json
2026-07-24 18:54:13,241 - vocalinux.ui.keyboard_backends - INFO - Detected desktop environment: x11
2026-07-24 18:54:13,241 - vocalinux.ui.keyboard_backends - INFO - Configured shortcut: ctrl+ctrl
2026-07-24 18:54:13,241 - vocalinux.ui.keyboard_backends - INFO - Configured mode: toggle
2026-07-24 18:54:13,241 - vocalinux.ui.keyboard_backends - INFO - Using pynput backend
2026-07-24 18:54:13,242 - vocalinux.ui.tray_indicator - INFO - All required resources validated successfully
2026-07-24 18:54:13,246 - vocalinux.suspend_handler - INFO - Suspend/resume handler initialized (logind D-Bus)
2026-07-24 18:54:13,246 - vocalinux.ui.keyboard_shortcuts - DEBUG - Cleared toggle callback
2026-07-24 18:54:13,246 - vocalinux.ui.keyboard_shortcuts - DEBUG - Cleared press callback
2026-07-24 18:54:13,246 - vocalinux.ui.keyboard_shortcuts - DEBUG - Cleared release callback
2026-07-24 18:54:13,246 - vocalinux.ui.tray_indicator - INFO - Setting up keyboard shortcuts with mode: toggle
2026-07-24 18:54:13,246 - vocalinux.ui.keyboard_shortcuts - INFO - Registered toggle callback for: Ctrl (either side)
2026-07-24 18:54:13,246 - vocalinux.ui.keyboard_shortcuts - INFO - Starting keyboard shortcut listener for: Ctrl (either side)
2026-07-24 18:54:13,246 - vocalinux.ui.keyboard_backends.pynput_backend - INFO - Starting pynput keyboard listener for shortcut: ctrl+ctrl (mode: toggle)
2026-07-24 18:54:13,247 - vocalinux.ui.keyboard_backends.pynput_backend - INFO - Pynput keyboard listener started successfully
2026-07-24 18:54:13,247 - vocalinux.ui.tray_indicator - INFO - Starting GTK main loop
2026-07-24 18:54:13,247 - vocalinux.ui.tray_indicator - INFO - Initializing system tray indicator
2026-07-24 18:54:13,247 - vocalinux.ui.tray_indicator - INFO - Using icon directory: /home/user87/.local/share/vocalinux/venv/lib/python3.13/site-packages/vocalinux/resources/icons/scalable
2026-07-24 18:54:13,248 - vocalinux.ui.tray_indicator - INFO - Icon directory exists: True
2026-07-24 18:54:13,248 - vocalinux.ui.tray_indicator - INFO - Available icon files: ['vocalinux-microphone-off.svg', 'vocalinux-microphone-process.svg', 'vocalinux-microphone.svg', 'vocalinux.svg']
2026-07-24 18:54:13,249 - vocalinux.ui.tray_indicator - INFO - Icon 'default' (/home/user87/.local/share/vocalinux/venv/lib/python3.13/site-packages/vocalinux/resources/icons/scalable/vocalinux-microphone-off.svg): exists
2026-07-24 18:54:13,249 - vocalinux.ui.tray_indicator - INFO - Icon 'active' (/home/user87/.local/share/vocalinux/venv/lib/python3.13/site-packages/vocalinux/resources/icons/scalable/vocalinux-microphone.svg): exists
2026-07-24 18:54:13,249 - vocalinux.ui.tray_indicator - INFO - Icon 'processing' (/home/user87/.local/share/vocalinux/venv/lib/python3.13/site-packages/vocalinux/resources/icons/scalable/vocalinux-microphone-process.svg): exists

(vocalinux:28962): libayatana-appindicator-WARNING **: 18:54:13.250: libayatana-appindicator is deprecated. Please use libayatana-appindicator-glib in newly written code.
```

### LINKED ISSUE #605: [FEATURE] Add trailing space after completed transcription
**Title:** Add trailing space after completed transcription

**Description:**

When dictating multiple transcriptions, Vocalinux currently inserts punctuation (such as a period) at the end of a transcription but does not insert a trailing space afterward.

Example:

Current behavior:

```
This is transcription one.This is transcription two.
```

Expected behavior:

```
This is transcription one. This is transcription two.
```

This is particularly noticeable when using push-to-talk or when dictation is performed in multiple segments. After each completed transcription, the user must manually press the space bar before continuing to dictate.

It would be helpful if Vocalinux automatically inserted a trailing space after a completed transcription segment, or provided this as an optional setting in Preferences.

* PR #464 fixed: "hello" [pause] "world" becoming helloworld
* My request: "hello." followed by another transcription becomes "hello. This" instead of "hello.This"

Benefits:

* More natural continuous dictation.
* Eliminates the need to manually press Space between dictated sentences.
* Matches the behavior of many speech-to-text applications.

Environment:

* Vocalinux version: 0.14.2
* Recognition engine: whisper.cpp
* OS: Debian 13
* Model: large-v3-turbo

### LINKED ISSUE #606: [BUG] IBus engine teardown raises TypeError: IBus.Object.destroy() takes exactly 1 argument (0 given)
**Version:** 0.14.2
**Affects:** every IBus-path user, all desktops — not compositor-specific

## Summary

`VocalinuxEngine.do_destroy()` passes `super().do_destroy` to `_handle_engine_destroy()`
as `super_destroy`. PyGObject binds that vfunc to the **GType**, not to the instance, so
calling it with no arguments leaves `IBus.Object.destroy` without its `self` and raises.

This fires on **every** engine teardown, i.e. after every dictation, and it is not merely
noisy: the exception escapes `_handle_engine_destroy()` before its `return`, so the two
state updates in the caller never execute.

## Traceback

```
2026-07-28 00:49:11,995 - __main__ - DEBUG - VocalinuxEngine instance destroyed
Traceback (most recent call last):
  File ".../vocalinux/text_injection/ibus_engine.py", line 864, in do_destroy
    VocalinuxEngine._active_instance = _handle_engine_destroy(
                                       ~~~~~~~~~~~~~~~~~~~~~~^
        VocalinuxEngine._active_instance,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<2 lines>...
        super_destroy,
        ^^^^^^^^^^^^^^
    )
    ^
  File ".../vocalinux/text_injection/ibus_engine.py", line 597, in _handle_engine_destroy
    super_destroy()
    ~~~~~~~~~~~~~^^
TypeError: IBus.Object.destroy() takes exactly 1 argument (0 given)
```

## Minimal reproduction

No Vocalinux needed — this is the raw PyGObject behaviour:

```python
import gi
gi.require_version("IBus", "1.0")
from gi.repository import IBus

class E(IBus.Engine):
    __gtype_name__ = "ProbeEngine"
    def probe(self):
        print(super().do_destroy)          # <bound method destroy of <GType ProbeEngine>>
        try:
            super().do_destroy()           # -> TypeError
        except TypeError as e:
            print("TypeError:", e)
        IBus.Object.destroy(self)          # -> OK
E().probe()
```

Output:

```
<bound method destroy of <GType ProbeEngine (...)>>
TypeError: IBus.Object.destroy() takes exactly 1 argument (0 given)
```

Note the binding target is the **GType**, not the instance — that is the whole bug.

## Why it matters

`do_destroy()` (v0.14.2, ibus_engine.py:857-871):

```python
VocalinuxEngine._active_instance = _handle_engine_destroy(...)   # line 864 — never runs
if was_active:
    VocalinuxEngine._focus_event.clear()                         # line 871 — never runs
```

Because `_handle_engine_destroy()` raises at line 597, neither assignment happens. After
every teardown the class is left with:

- `_active_instance` still pointing at a **destroyed** engine, and
- `_focus_event` still **set**.

`_ibus_ping_status()` (the inject-socket PING readiness probe added in #523) derives its
answer from exactly those two values, so it can report `FOCUSED` — "bound to a client
context, safe to commit" — on the strength of a destroyed instance. In practice
`do_focus_in()` reassigns `_active_instance` on the next activation, which is likely why
this has gone unnoticed; the stale window is between dictations. I have not managed to
attribute a specific dropped commit to it, so I am reporting the state corruption as
observed fact and the commit risk as inference.

## Suggested fix

Two independent changes; the second is worth having regardless of the first.

**1. Bind the parent destroy to the instance** (ibus_engine.py:862):

```python
super_destroy: Optional[Callable[[], None]] = None
if IBUS_AVAILABLE:
    parent_destroy = super().do_destroy

    def super_destroy() -> None:
        # PyGObject binds this vfunc to the GType rather than to the instance,
        # so calling it with no arguments raises "IBus.Object.destroy() takes
        # exactly 1 argument (0 given)". Retry with the instance explicitly.
        try:
            parent_destroy()
        except TypeError:
            IBus.Object.destroy(self)
```

The `try`/`except` rather than calling `IBus.Object.destroy(self)` outright keeps working
if a future PyGObject binds the vfunc correctly.

**2. Never let a failing parent destroy skip teardown** (ibus_engine.py:596-597):

```python
if ibus_available and super_destroy is not None:
    try:
        super_destroy()
    except Exception:
        logger.exception("Parent engine destroy failed; continuing teardown")

return next_active_instance
```

The caller assigns this function's return value to `_active_instance` and clears the focus
event, so no failure inside it should be able to skip that.

## Verification

With both applied, teardown state is correct even when the parent destroy fails:

```python
def boom(): raise TypeError("IBus.Object.destroy() takes exactly 1 argument (0 given)")
sentinel = object()
assert _handle_engine_destroy(sentinel, sentinel, True, boom) is None   # passes
```

And in normal operation the traceback is gone — engine log across a full dictation cycle:

```
01:00:09,864 - VocalinuxEngine instance created
01:00:09,864 - VocalinuxEngine focus in
01:00:09,864 - VocalinuxEngine enabled - ready for text injection
01:00:10,070 - Injecting text: But some more dictations, la la la la la la...
01:00:10,070 - Text injected via IBus: 'But some more dictat...' (43 chars)
01:00:10,074 - VocalinuxEngine focus out
01:00:10,074 - VocalinuxEngine disabled (keeping instance for injection)
01:00:10,074 - VocalinuxEngine instance destroyed
```

## Environment

| | |
|---|---|
| Vocalinux | 0.14.2 |
| IBus | 1.5.34 |
| PyGObject | 3.56.3 |
| Python | 3.14.6 |
| Compositor | Hyprland 0.56.1 (Wayland) |

The reproduction above depends only on IBus + PyGObject, so the compositor is incidental.

### LINKED ISSUE #607: [FEATURE] Allow IBus on wlroots compositors when ibus-wayland (input-method-v2) is running
**Version:** 0.14.2
**Relates to:** #485, #486 (the denylist), #476 / #402 (selectable backends)

## Summary

`_IBUS_UNBRIDGED_COMPOSITORS` (text_injector.py:197) permanently disables IBus injection on
`cosmic`, `sway`, `hyprland`, `wayfire`, `river`, `niri`, `labwc` and `weston`. The premise
in #485 was that on these compositors nothing carries IBus commits to native Wayland
clients, so `commit_text()` succeeds while the text is silently dropped.

That premise is only true when nothing is bridging the compositor to `ibus-daemon`. Since
**IBus 1.5.32**, IBus ships `ibus-wayland`, which implements `zwp_input_method_v2` — exactly
the missing link. With it running, IBus reaches native Wayland clients on these compositors,
including Electron/Chromium, and the denylist blocks a path that works.

The denylist is a static guess about the environment where a cheap runtime fact is
available. I would like to propose replacing the guess with a probe.

## Evidence

On Hyprland 0.56.1 with IBus 1.5.34:

```console
$ wayland-info | grep -E 'text_input|input_method'
interface: 'zwp_text_input_manager_v1',    version: 1, name: 20
interface: 'zwp_text_input_manager_v3',    version: 1, name: 21
interface: 'zwp_input_method_manager_v2',  version: 1, name: 28

$ strings /usr/lib/ibus/ibus-wayland | grep zwp_input_method
zwp_input_method_keyboard_grab_v2
zwp_input_method_manager_v2
zwp_input_method_v1
zwp_input_method_context_v1
zwp_input_method_v2
```

The compositor offers `zwp_input_method_manager_v2`; IBus implements it. Nothing was
connecting the two, which is what makes the denylist look correct.

With `ibus-wayland` started (and a global engine set — it logs `No global engine` and
cannot route commits without one), dictation into Electron apps works through IBus:

```
VocalinuxEngine enabled - ready for text injection
Injecting text: We are testing dictation again now ...
Text injected via IBus: 'We are testing dicta...' (70 chars)
```

303 chars committed in ~200 ms, versus 16.77 s for 415 chars on the `ydotool type` path
(`-d 20 -H 20` defaults = 40 ms/char). Verified in VSCodium, Claude Desktop, a third
Electron app, and kitty.

One caveat worth documenting either way: Chromium only opens a `zwp_text_input_v3` seat when
launched with `--enable-wayland-ime`. Without that flag the bridge runs and the app still
receives nothing — which may be part of why IBus looked unusable with Electron.

## Proposed change

Add a live probe and let it override the denylist:

```python
@staticmethod
def _ibus_wayland_bridge_running() -> bool:
    """Whether IBus' zwp_input_method_v2 bridge (ibus-wayland) is running."""
    try:
        result = subprocess.run(
            ["pgrep", "-x", "ibus-wayland"], capture_output=True, timeout=2
        )
        return result.returncode == 0
    except (subprocess.SubprocessError, FileNotFoundError):
        return False

def _wayland_compositor_bridges_ibus(self) -> bool:
    if self.environment != DesktopEnvironment.WAYLAND:
        return True
    if self._ibus_wayland_bridge_running():
        logger.info(
            "ibus-wayland input-method-v2 bridge is running; IBus can reach "
            "native Wayland clients on this compositor"
        )
        return True
    desktop = " ".join(
        os.environ.get(var, "")
        for var in ("XDG_CURRENT_DESKTOP", "XDG_SESSION_DESKTOP", "DESKTOP_SESSION")
    ).lower()
    return not any(name in desktop for name in self._IBUS_UNBRIDGED_COMPOSITORS)
```

Properties worth noting:

- **Strictly additive.** With no bridge running, behaviour is byte-for-byte today's. Users
  who have not set `ibus-wayland` up are unaffected.
- **Self-disabling.** If the bridge dies or is never started, the next `TextInjector`
  construction falls back to the denylist automatically. Because the probe is a live fact
  rather than a config flag, there is no way to assert a bridge that isn't there.
- **No new config surface.**

The cost is one `pgrep` per `TextInjector` construction, not per injection.

## Optional companion: `VOCALINUX_FORCE_BACKEND`

While testing this I also added an env override — `ibus` / `wtype` / `ydotool` / `auto`
(default) — which made A/B-ing the two paths practical without editing code, and doubles as
an escape hatch when a backend misbehaves. It overlaps with #476 and #402, so I mention it
only in case a minimal env-var form is useful before the full settings-UI work lands. Happy
to split it into its own PR or drop it.

## Why not just document "run ibus-wayland"?

Documenting it alone doesn't help: the denylist would still refuse IBus even with the bridge
running, so a user who sets it up correctly still lands on `wtype`/`ydotool` with no
indication why. The probe is what makes the setup actionable.

## Environment

| | |
|---|---|
| Vocalinux | 0.14.2 |
| IBus | 1.5.34 (`ibus-wayland` present) |
| Compositor | Hyprland 0.56.1 |
| Session | Wayland, `GTK_IM_MODULE`/`QT_IM_MODULE` unset, `XMODIFIERS=@im=ibus` |

I have this running locally as a patch and am happy to open a PR if the approach looks
reasonable.

jatinkrmalik:
## Description

Prepares **v0.15.0** for everything on `main` since `v0.14.2`.

**Rebased onto latest `main` again** — includes **#618** (sidebar dictation footer), **#619** (Custom Shortcut Record/Set restore), **#617** (English India → Whisper `en`), plus #616 / #599 and earlier fold-ins.

### Why 0.15.0 (not 1.0)

`0.14.1` and `0.14.2` were already stable. This delta is a minor feature set. AppImage just landed, so it belongs in 0.15 before anyone treats a build as a forever 1.0 baseline.

### In this PR

- Version bump to `0.15.0` across app, web, AUR, Flatpak AppStream
- Production/Stable classifier; release-notes writing rules in `docs/RELEASE_PROCESS.md`
- Docs + website updated for the full 0.15 delta

### Smoke evidence (tip of main + local AppImage)

Short recap video (settings + AppImage dictation via IBus):

[vocalinux-smoke-recap-playable.mp4](https://cursor.com/agents/bc-c016795c-af84-42c5-98d1-1c5e7b174f9d/artifacts?path=%2Fopt%2Fcursor%2Fartifacts%2Fvocalinux-smoke-recap-playable.mp4)

Screenshots:

Settings Dictation page
whisper.cpp dictation
AppImage dictation

Social / Product Hunt drafts stay outside git (maintainer artifacts only).

## Checklist

- [x] Rebased onto latest `main` (includes #617–#619)
- [x] Notes include settings footer, shortcut UI restore, en-in Whisper mapping, languages, Bluetooth mic fix
- [x] CI green (re-check after this push)
- [x] Local install smoke (whisper.cpp / VOSK / remote API / IBus)
- [x] Local AppImage build + dictation smoke
- [ ] Tag **after merge** only

---

## Draft GitHub Release Notes

# Vocalinux v0.15.0

Vocalinux 0.15.0 is a stable minor release on top of the 0.14 packaging line. Settings move to a searchable sidebar with dictation controls in the footer, GitHub Releases include AppImages, the speech-language catalog grows a lot (including Hungarian), continuous dictation gets capitalization and spacing fixes, and there are new controls for unloading the model and picking a Vulkan GPU. Wayland IBus also improves when `ibus-wayland` is present. Selecting a Bluetooth headset mic no longer crashes the process during PortAudio probing.

---

## Highlights

| Area | Description |
|------|-------------|
| Settings | Sidebar navigation with search; status / mic level / Test Dictation / Close stay in the sidebar footer |
| AppImage | x86_64 and aarch64 builds attached to the GitHub Release |
| Languages | ~33 selectable speech languages plus Auto-detect (incl. Hungarian); VOSK only lists languages with official models |
| Dictation polish | Capitalize after `.` / `!` / `?`; leave a trailing space after each completed utterance |
| Auto-pause / keep-alive | Unload the model while configured apps run, or after an idle timeout |
| Vulkan devices | Prefer a discrete GPU when one exists; override under Advanced settings |
| ibus-wayland | Prefer IBus on previously “unbridged” compositors when `ibus-wayland` is running |
| CLI | `vocalinux --version` prints the package version and exits |
| Bluetooth mics | Capture open/probe path no longer heap-corrupts on SCO/HFP devices |

---

## New Features

- Settings: searchable sidebar instead of seven notebook tabs. (#601 by @jatinkrmalik)
- Settings: move recognition status, mic level, Test Dictation, and Close into the always-visible sidebar footer. (#618 by @jatinkrmalik)
- Packaging: AppImage for x86_64 and aarch64, plus smoke-test packaging fixes. (#573, #602 by @jatinkrmalik)
- Languages: expand the speech catalog (Hungarian and many more); CLI `--language` choices come from the same list as Settings; wire official Alphacephei VOSK models where they exist and hide Whisper-only languages in the VOSK dropdown. (#616 by @jatinkrmalik, fixes #565)
- Dictation: capitalize at the start and after sentence-ending punctuation. (#554 by @jatinkrmalik, closes #553)
- Dictation: append a trailing space after each completed transcription so the next session does not glue on. (#608 by @jatinkrmalik, fixes #605)
- Auto-pause competing apps and idle model keep-alive unload. (#592 by @jatinkrmalik, closes #445, #591)
- Vulkan: auto-select a discrete GPU and allow a manual device in Advanced settings. (#590 by @jatinkrmalik, closes #589)
- Wayland: use IBus when `ibus-wayland` is running, including on compositors previously treated as unbridged. (#614 by @eiseleb47, closes #607)
- CLI: `vocalinux --version`. (#563 by @AnayGarodia, closes #555)
- Website: marketing site redesign and languages page with honest per-engine badges. (#582, #616 by @jatinkrmalik)

---

## Bug Fixes

- Settings: restore Custom Shortcut entry / Record / Set controls after the sidebar settings refactor. (#619 by @jatinkrmalik)
- Languages: map English (India) (`en-in`) to Whisper code `en` for whisper.cpp / Whisper / remote API (VOSK still uses `en-in`). (#617 by @jatinkrmalik)
- Audio: stop Bluetooth mic probing from corrupting the heap. One PortAudio open per capture session, stop-before-close, and no stereo probe on mono-only devices (fixes crashes on SCO/HFP headsets such as PipeWire “Bluetooth internal capture”). (#599 by @jatinkrmalik, fixes #567 reported by @HaleTom)
- IBus: keep engine teardown correct when parent `do_destroy` fails. (#613 by @eiseleb47, fixes #606)
- Settings UI: flatten info notices so helper text matches the rest of the dialog. (#615 by @jatinkrmalik)
- KDE Plasma Wayland: skip unbridged IBus when `ibus-wayland` is not present so injection does not fail quietly. (#577 by @jatinkrmalik, fixes #574)
- xdotool: keep input focus after injection. (#564 by @AnayGarodia, fixes #549)
- Uninstall: remove the IBus data dir, stop the app via PID file, and remove `~/.local/bin` launchers reliably. (#597 by @jatinkrmalik, #569 by @webenefits)
- Installer: install both libgirepository 1.0 and 2.0 when the distro has them. (#583 by @jatinkrmalik, fixes #571)
- AUR: depend on virtual `python-pywhispercpp`; move clipboard and wtype tools to optdepends. (#579 by @Asseel-Naji, #586 by @jatinkrmalik)
- UI: first-run dialog response without `Gtk.Dialog.do_response`. (#580 by @jatinkrmalik, fixes #566)
- Vosk: add Italian and English-India rows to the medium/large model tables. (#551 by @AnayGarodia, fixes #550)
- Docs: fix ydotool service setup guidance. (#560 by @AnayGarodia, fixes #557)

### Docs / maintenance (lower priority)

- Multi-distro tray icon FAQ. (#584)
- robots.txt no longer blocks indexable pages. (#610)
- Website CI lint / action warning cleanup. (#611)
- README badge layout tweaks. (#542, #544, #545)
- Next.js dependency bump. (#587)

---

## Thanks

Thank you **@HaleTom** for reporting the Bluetooth mic malloc crash (#567).

Thank you **@eiseleb47** for the IBus teardown fix and for teaching Vocalinux about `ibus-wayland` (#613, #614).

Thank you **@AnayGarodia** for `--version`, the xdotool focus fix, the Vosk model table entries, and the ydotool docs correction (#563, #564, #551, #560).

Thank you **@webenefits** for making uninstall remove the `~/.local/bin` wrappers (#569).

Thank you **@Asseel-Naji** for the AUR `python-pywhispercpp` dependency fix (#579).

Thanks to everyone who filed the linked issues (including #565 for Hungarian) and tested builds on the way to 0.15.

---

### Recommended (install.sh)

```bash
curl -fsSL https://raw.githubusercontent.com/jatinkrmalik/vocalinux/main/install.sh -o /tmp/vl.sh && bash /tmp/vl.sh
```

### AppImage

```bash
chmod +x Vocalinux-0.15.0-x86_64.AppImage
./Vocalinux-0.15.0-x86_64.AppImage
```

Host text-injection tools (`xdotool` / `wtype` / `ydotool`) are still required. FUSE is needed to run the AppImage on most hosts.

### Arch / PyPI / Flatpak

`yay -S vocalinux` · `pip install -U vocalinux` · local Flatpak under `packaging/flatpak/` (Flathub may trail this tag)

**Full changelog:** https://github.com/jatinkrmalik/vocalinux/compare/v0.14.2...v0.15.0

<div><a href="https://cursor.com/agents/bc-c016795c-af84-42c5-98d1-1c5e7b174f9d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-c016795c-af84-42c5-98d1-1c5e7b174f9d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


jatinkrmalik:
Tested on both local and cloud vm! 

image

Issue #523, opened 2026-07-15
[BUG] Vocalinux reported "Text injection completed successfully" while the text never reached the app


BlackEyedHatMan:
Title: First dictation of a session silently drops text on GNOME Wayland (scoped IBus injection, bare xkb engine) — logs report success

Environment: Vocalinux v0.14.0-beta, Ubuntu GNOME on Wayland, baseline IBus engine xkb:gb::eng (bare layout, so the scoped-injection path from #500/#506 is used), whisper.cpp backend.

Symptom: On the very first dictation after login, transcription succeeds and the log shows Starting IBus text injection → Text injection completed successfully, but no text appears in the focused app (GNOME Text Editor). Every  subsequent dictation in the same session works fine.

Diagnosis done: Traced the IBus bus with dbus-monitor --address "$(ibus address)" while injecting through the engine's inject.sock directly. Once the vocalinux engine has been activated at least once in the session, the full chain works  — SetGlobalEngine → FocusIn → CommitText routed to gnome-shell's input context → the app echoes the committed text back via SetSurroundingText (confirmed inserted), even with the immediate engine restore. GNOME does not revert the engine switch. The failure only occurred on the session's first-ever activation of the engine (a trace of the failing case wasn't captured, so the exact drop point is inferred).

Hypothesis: The background warmup (TextInjector._initialize_ibus_in_background → prepare_engine) starts the engine process and registers the component but intentionally never activates the engine. So the session's first real injection is also the first-ever SetGlobalEngine vocalinux, and the commit is emitted ~200ms later while gnome-shell/mutter's input-method state for the newly-seen engine is still settling — the commit is dropped between compositor and client. 
inject_text() reports success because success is measured at commit emission, not delivery.

Possible fixes:
1. Have the warmup perform one scoped activate-and-restore cycle (switch_engine("vocalinux") → restore) at startup, so the first real injection never hits the cold path.
2. In the engine, wait for do_focus_in on the fresh engine instance before committing, instead of relying on the fixed sleep(0.2) in switch_engine().
3. Where the client provides surrounding text, verify the committed string appears in do_set_surrounding_text and retry once on mismatch — this would also fix the misleading "completed successfully" log.

## 💻 Environment
- OS: Ubuntu Ubuntu 26.04 LTS
- Python Version: 3.14.4
- Vocalinux Version: v0.14.0-beta
- Display Server: Wayland
- Speech Engine:  Whisper
- Model Size: small

[...]

PR #613, opened 2026-07-28
fix(ibus): keep engine teardown correct when parent destroy fails

### LINKED ISSUE #606: [BUG] IBus engine teardown raises TypeError: IBus.Object.destroy() takes exactly 1 argument (0 given)

eiseleb47:
Fixes #606.

## What

`do_destroy` passed `super().do_destroy` into `_handle_engine_destroy`. PyGObject binds that vfunc to the **GType** rather than to the instance, so calling it with no arguments raises:

```
TypeError: IBus.Object.destroy() takes exactly 1 argument (0 given)
```

This fired on every engine teardown, i.e. after every dictation.

## Why it matters beyond the noise

The exception escaped `_handle_engine_destroy` before its `return`, so neither statement in the caller ran:

```python
VocalinuxEngine._active_instance = _handle_engine_destroy(...)   # skipped
if was_active:
    VocalinuxEngine._focus_event.clear()                         # skipped
```

That left a destroyed engine registered as active with the focus event still set — the two values `_ibus_ping_status` (#523) reads to decide whether it is safe to commit. `do_focus_in` reassigns `_active_instance` on the next activation, which is likely why this went unnoticed; the stale window is between dictations. I could not attribute a specific dropped commit to it, so I am reporting the state corruption as fact and the commit risk as inference.

## Changes

1. **Bind the parent destroy to the instance.** Falls back to `IBus.Object.destroy(self)` only when the no-arg call raises `TypeError`, so this keeps working if a future PyGObject binds the vfunc correctly.
2. **Guard the parent destroy call**, so no failure there can skip teardown. The caller depends on the return value, so nothing inside should be able to bypass it.

## Reproduction

Independent of Vocalinux — raw IBus + PyGObject:

```python
import gi
gi.require_version("IBus", "1.0")
from gi.repository import IBus

class E(IBus.Engine):
    __gtype_name__ = "ProbeEngine"
    def probe(self):
        print(super().do_destroy)   # <bound method destroy of <GType ProbeEngine>>
        super().do_destroy()        # TypeError
E().probe()
```

Note the binding target is the GType, not the instance.

## Tests

Two added, covering teardown with a raising parent destroy (both the "is active" and "is a different instance" paths). Verified no new flake8 findings and no test failures unique to this branch against `main`.

## Environment

Vocalinux 0.14.2 · IBus 1.5.34 · PyGObject 3.56.3 · Python 3.14.6 · Hyprland 0.56.1. The reproduction depends only on IBus + PyGObject, so the compositor is incidental.

### DIFF
--- a/src/vocalinux/text_injection/ibus_engine.py
+++ b/src/vocalinux/text_injection/ibus_engine.py
@@ -601,7 +601,13 @@ def _handle_engine_destroy(
         next_active_instance = None
 
     if ibus_available and super_destroy is not None:
-        super_destroy()
+        # The caller assigns our return value to _active_instance and clears the
+        # focus event, so a failing parent destroy must not skip that teardown:
+        # otherwise a destroyed engine stays registered as active with focus set.
+        try:
+            super_destroy()
+        except Exception:
+            logger.exception("Parent engine destroy failed; continuing teardown")
 
     return next_active_instance
 
@@ -866,7 +872,20 @@ class VocalinuxEngine(IBus.Engine if IBUS_AVAILABLE else object):
         logger.debug("VocalinuxEngine instance destroyed")
         super_destroy: Optional[Callable[[], None]] = None
         if IBUS_AVAILABLE:
-            super_destroy = super().do_destroy
+            parent_destroy = super().do_destroy
+
+            def destroy_via_parent() -> None:
+                # PyGObject binds this vfunc to the GType rather than to the
+                # instance, so calling it with no arguments raises
+                # "IBus.Object.destroy() takes exactly 1 argument (0 given)".
+                # Retry with the instance explicitly. The try/except keeps this
+                # working if a future PyGObject binds the vfunc correctly.
+                try:
+                    parent_destroy()
+                except TypeError:
+                    IBus.Object.destroy(self)
+
+            super_destroy = destroy_via_parent
         was_active = VocalinuxEngine._active_instance is self
         VocalinuxEngine._active_instance = _handle_engine_destroy(
             VocalinuxEngine._active_instance,
--- a/src/vocalinux/text_injection/ibus_engine.py
+++ b/src/vocalinux/text_injection/ibus_engine.py
@@ -14,6 +14,7 @@ This is the preferred method for Wayland environments as it works universally
 without requiring compositor-specific protocols.
 """
 
+import functools
 import logging
 import os
 import signal
@@ -589,6 +590,23 @@ def restore_xkb_layout(layout: str, variant: str = "", option: str = "") -> bool
     return False
 
 
+def _invoke_parent_destroy(parent_destroy: Callable[[], None], instance: object) -> None:
+    """Call the parent engine destroy, working around PyGObject's GType binding.
+
+    PyGObject binds the ``do_destroy`` vfunc to the GType rather than to the
+    instance, so ``super().do_destroy()`` raises
+    ``TypeError: IBus.Object.destroy() takes exactly 1 argument (0 given)``.
+    Retry with the instance explicitly.
+
+    The TypeError path rather than an unconditional ``IBus.Object.destroy``
+    keeps this working if a future PyGObject binds the vfunc correctly.
+    """
+    try:
+        parent_destroy()
+    except TypeError:
+        IBus.Object.destroy(instance)
+
+
 def _handle_engine_destroy(
     active_instance: Optional["VocalinuxEngine"],
     current_instance: object,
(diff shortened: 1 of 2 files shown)


jatinkrmalik (approved):
Looks good. The instance-bound parent destroy plus the teardown guard fixes the #606 state leak cleanly, and the helper + tests pin the contract well. LGTM.

Issue #476, opened 2026-06-19
[FEATURE] Let users pick the text-injection backend (e.g. force wtype instead of IBus)


waldemar-p:
## 🚀 Feature Description
Let users pick the text-injection backend — e.g. force `wtype` instead of IBus on Wayland. Right now there's no way to tell vocalinux which one to use.

## 💡 Problem Statement
On Wayland, vocalinux always goes for IBus when it's available, and there's no way to tell it not to. That's a problem on compositors where IBus doesn't actually deliver text to native apps.

My case: Pop!_OS with COSMIC. IBus injection works in XWayland stuff (browser, vocalinux's own field) but nothing shows up in native Wayland apps like the COSMIC editor or Zed. `wtype` works perfectly on the same machine, German umlauts and all. But I can't get vocalinux onto wtype, and there's no override to make it happen.

## 📋 Proposed Solution
An explicit backend override like an `injection_backend` key in the existing `config.json`. When set to a non-IBus backend, vocalinux would skip the IBus path entirely and not start ibus-daemon. Left unset, everything stays exactly as it is now.

## 🔄 Alternatives Considered
- Removing IBus — not an option, there are other dependencies.
- Killing `ibus-daemon` first — doesn't help, vocalinux just restarts it once it picks the IBus path.

## ✅ Acceptance Criteria
- [ ] A config to select injection backend.
- [ ] When set to a non-IBus backend, vocalinux doesn't enter the IBus path or start ibus-daemon.
- [ ] When unset, current auto-detection behavior is unchanged.

jatinkrmalik:
@waldemar-p We had a PR #402 by @sabbari that started this work, but seems like it was abandoned. Let me see if I can pick it up over the weekend just in time for our next version bump release.

luginf:
I have the same problem. Not on wayland, but on X11 as well, ibus is not working out of the box. The only workaround I found to use xdotool instead is to create a script like that:


~/.local/bin/vocalinux.sh
```
#!/bin/bash
export GTK_IM_MODULE=""
export QT_IM_MODULE=""
export XMODIFIERS=""
exec ~/.local/share/vocalinux/venv/bin/vocalinux "$@"
```

and so I start it with vocalinux.sh instead of just vocalinux.sh

Maybe just an option to disable the ibus backend would be sufficent

PR #614, opened 2026-07-28
feat(text-injection): use IBus on unbridged compositors when ibus-wayland runs

### LINKED ISSUE #607: [FEATURE] Allow IBus on wlroots compositors when ibus-wayland (input-method-v2) is running

eiseleb47:
Closes #607.

## What

`_IBUS_UNBRIDGED_COMPOSITORS` assumes nothing sits between the compositor and ibus-daemon. Since **IBus 1.5.32** that is no longer safe: IBus ships `ibus-wayland`, which implements `zwp_input_method_v2` and relays commits to native Wayland clients speaking text-input-v3. Every compositor on the denylist exposes `zwp_input_method_manager_v2`, so with the bridge running they *are* bridged — and skipping IBus drops to a slower path for no reason.

This adds a live `pgrep -x ibus-wayland` probe and lets it override the denylist.

## Scope

The probe is checked **only inside the denylist branch**. GNOME and the KDE VirtualKeyboard path (#574) are untouched — there is a test asserting the probe is never even consulted for bridged desktops.

Properties:

- **Strictly additive.** With no bridge running, behaviour is byte-for-byte what it is today.
- **Self-disabling.** A live probe rather than configuration, so it degrades back to the denylist by itself if the bridge dies or was never started. There is no way to assert a bridge that isn't there.
- **Cost** is one `pgrep` per `TextInjector` construction, not per injection.

## Evidence

Hyprland 0.56.1, IBus 1.5.34:

```console
$ wayland-info | grep -E 'text_input|input_method'
interface: 'zwp_text_input_manager_v1',    version: 1, name: 20
interface: 'zwp_text_input_manager_v3',    version: 1, name: 21
interface: 'zwp_input_method_manager_v2',  version: 1, name: 28

$ strings /usr/lib/ibus/ibus-wayland | grep zwp_input_method
zwp_input_method_manager_v2
zwp_input_method_v2
...
```

With the bridge started, dictation commits through IBus in VSCodium, Claude Desktop, another Electron 41 app and kitty — 303 chars in ~200 ms, versus 16.8 s for 415 chars via `ydotool type` (`-d 20 -H 20` defaults = 40 ms/char).

**Caveat worth documenting either way:** Chromium only opens a `zwp_text_input_v3` seat when launched with `--enable-wayland-ime`. Without it the bridge runs and the app still receives nothing — possibly part of why IBus has looked unusable with Electron.

## VOCALINUX_FORCE_BACKEND

Folded in per your note on #607. Accepts `ibus` / `wtype` / `ydotool` / `auto`; unknown values fall back to `auto` with a warning rather than silently pinning a backend.

The motivation is that autodetection has to *infer* whether IBus commits reach the focused app and cannot verify it — `commit_text()` reports success even when the text is dropped — so an escape hatch is useful when the inference is wrong, and it makes the two paths A/B-testable without editing code.

This overlaps #476 / #402. **Happy to drop it from this PR** if you would rather it follow the design there — the probe stands alone and I would rather that not wait on it.

## Two existing tests changed

Both asserted the unbridged outcome without pinning the probe, so both were implicitly host-dependent once a probe exists:

- `test_cosmic_skips_ibus_and_uses_wtype` blanket-mocks `subprocess.run` to `returncode=0`, which the probe reads as "bridge found".
- `test_unbridged_wayland_skips_ibus_when_engine_is_xkb` shells out for real, so it depended on whether the machine running the suite had the bridge up.

Both now state the "no bridge" precondition explicitly. Their intent is unchanged.

## Tests

10 added: denylisted compositors with and without the bridge, probe not consulted for bridged desktops, probe argv, probe resilience to missing/hung `pgrep`, and the `VOCALINUX_FORCE_BACKEND` parsing matrix. No new flake8 findings and no test failures unique to this branch against `main`.

### DIFF
--- a/src/vocalinux/text_injection/text_injector.py
+++ b/src/vocalinux/text_injection/text_injector.py
@@ -284,6 +284,29 @@ class TextInjector:
         )
         return False
 
+    @staticmethod
+    def _ibus_wayland_bridge_running() -> bool:
+        """Whether IBus' zwp_input_method_v2 bridge (``ibus-wayland``) is running.
+
+        The ``_IBUS_UNBRIDGED_COMPOSITORS`` denylist assumes nothing sits between
+        the compositor and ibus-daemon. Since IBus 1.5.32 the ``ibus-wayland``
+        helper implements ``zwp_input_method_v2``, so on a compositor that
+        exposes ``zwp_input_method_manager_v2`` (the wlroots/smithay ones on the
+        denylist all do) it relays commits to native Wayland clients speaking
+        text-input-v3. When it is running, those compositors are bridged.
+
+        Mirrors ``is_ibus_daemon_running()`` in ibus_engine.py.
+        """
+        try:
+            result = subprocess.run(
+                ["pgrep", "-x", "ibus-wayland"],
+                capture_output=True,
+                timeout=2,
+            )
+            return result.returncode == 0
+        except (subprocess.SubprocessError, FileNotFoundError):
+            return False
+
     def _wayland_compositor_bridges_ibus(self) -> bool:
         """Whether native Wayland clients receive IBus commits on this compositor.
 
@@ -293,6 +316,12 @@ class TextInjector:
         those we must use a virtual-keyboard tool (wtype/ydotool) instead. We use
         a denylist rather than an allowlist so that unrecognised desktops keep the
         previous IBus-preferred behaviour.
+
+        A denylisted compositor is still bridged when ``ibus-wayland`` is running,
+        since that supplies exactly the input-method-v2 relay those compositors
+        lack (issue #607). The check is a live probe rather than configuration,
+        so this degrades back to the denylist on its own if the bridge dies or
+        was never started.
 
         On KDE Plasma Wayland, bridging also requires KWin VirtualKeyboard to be
         enabled (issue #574); otherwise commit_text succeeds at the IBus layer
@@ -306,6 +335,14 @@ class TextInjector:
             for var in ("XDG_CURRENT_DESKTOP", "XDG_SESSION_DESKTOP", "DESKTOP_SESSION")
         ).lower()
         if any(name in desktop for name in self._IBUS_UNBRIDGED_COMPOSITORS):
+            if self._ibus_wayland_bridge_running():
+                logger.info(
+                    "Compositor '%s' is on the unbridged list, but the ibus-wayland "
+                    "input-method-v2 bridge is running; IBus can reach native "
+                    "Wayland clients.",
+                    os.environ.get("XDG_CURRENT_DESKTOP", "unknown"),
+                )
+                return True
             return False
         if _is_kde_plasma_session():
             return self._kde_virtual_keyboard_enabled()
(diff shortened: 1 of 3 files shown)


jatinkrmalik (approved):
Nice. Keeping the `ibus-wayland` probe inside the denylist branch is the right shape so the KDE path stays alone, and leaving force-backend in here is fine. Tests cover the host-dependent cases well. LGTM.

PR #621, opened 2026-07-30
Fix missing KDE tray icon by preferring Ayatana AppIndicator

kacperpaczos:
## Problem

On KDE Plasma the Vocalinux tray icon never shows up, even though the app starts and logs nothing wrong. Reproduced on a fresh install on Fedora 44 (KDE Plasma 6.7.3, Wayland).

Root cause: Fedora and Arch installs pull `libappindicator-gtk3`, Canonical's original AppIndicator library, which hasn't had a real release since around 2013. `AppIndicator3.Indicator.new_with_path()` and `set_status(ACTIVE)` both run without raising an exception, so nothing gets logged, but the item is never registered with KDE's `StatusNotifierWatcher`. I confirmed this directly by querying `RegisteredStatusNotifierItems` over D-Bus: the process shows up with open D-Bus connections but no entry in the watcher's item list.

The actively maintained Ayatana fork does register correctly, and it's already what the Debian, Ubuntu, openSUSE and Gentoo branches of `install.sh` install. Fedora and Arch were the two distros still defaulting to the legacy package.

## Fix

- `src/vocalinux/ui/tray_indicator.py`: reorder the `gi.repository` import to try `AyatanaAppIndicator3` (and the lowercase `AyatanaAppindicator3` variant) first, falling back to the legacy `AppIndicator3` typelib only if neither Ayatana variant is installed. Both are still supported, so this doesn't break systems where only the legacy package is present.
- `install.sh`: added a small `install_preferred_appindicator()` helper used by the Fedora and Arch branches. It tries the Ayatana package first and falls back to the legacy one if unavailable (e.g. RHEL/CentOS derivatives without EPEL), matching the pattern the Ubuntu and openSUSE branches already use.

## Testing

- Verified on Fedora 44 / KDE Plasma 6.7.3 / Wayland: before the fix, `RegisteredStatusNotifierItems` never included Vocalinux; after installing `libayatana-appindicator-gtk3` and applying the code change, the item registers and the tray icon appears (both immediately and after a `plasmashell` restart, without logging out).
- `tests/test_tray_indicator.py` and `tests/test_tray_indicator_ext.py` pass unchanged.
- `bash -n install.sh` and `shellcheck -x install.sh` show no new issues from the added function.
- `flake8` on the changed Python file shows the same pre-existing warnings as before (none introduced by this change).


jatinkrmalik (changes requested):
Can you please address the bot feedback above?

kacperpaczos:
Addressed the bot's feedback: `install_preferred_appindicator()` now only skips straight to a no-op when Ayatana is already installed, and tries installing Ayatana first even if the legacy `libappindicator-gtk3` is already present. It only falls back to "legacy is already there" (with the KDE warning) if the Ayatana install itself fails. That way re-running install.sh on an existing Fedora/Arch setup picks up the fix instead of silently keeping the broken legacy package.

All checks are green, including codecov/patch and codecov/project.

jatinkrmalik (approved):

jatinkrmalik:
LGTM

PR #635, opened 2026-08-03
fix(installer): gate util-linux-extra to Ubuntu 24.04+

### LINKED ISSUE #526: [BUG] install.sh fails on Linux Mint 21.3 (Ubuntu 22.04 base): util-linux-extra not found
## 🐛 Bug Description
install.sh (both the standard and --dev modes) fails on Ubuntu 22.04-based systems because it requires the util-linux-extra package, which does not exist on this release.

## 📋 Steps to Reproduce
1. On Linux Mint 21.3 (Ubuntu 22.04 base), run: curl -fsSL raw.githubusercontent.com/jatinkrmalik/vocalinux/main/install.sh -o /tmp/vl.sh && bash /tmp/vl.sh --interactive
2. Same result with: git clone the repo, then ./install.sh --dev
3. See error

## ✅ Expected Behavior
The installer should complete dependency installation on any Ubuntu 22.04+ / Debian 11+ base, as stated in the README's supported OS list.

## ❌ Actual Behavior
E: Unable to locate package util-linux-extra
[ERROR] Failed to install dependencies

util-linux-extra only exists from Ubuntu 24.04 / Debian 12 onward, following a package split starting at util-linux 2.38. Ubuntu 22.04 ships util-linux 2.37.2 without this split, so the package genuinely isn't available in the repositories for this release.

## 📸 Screenshots
Not applicable, terminal error only.

## 💻 Environment
- **OS**: Linux Mint 21.3 (Cinnamon), Ubuntu 22.04 base
- **Python Version**: 3.10
- **Vocalinux Version**: 0.14.0-beta
- **Display Server**: X11
- **Speech Engine**: whisper.cpp
- **Model Size**: small

## 📋 Debug Logs
Not applicable, the failure happens in install.sh before the application itself runs.

## 📝 Additional Context
Workaround: installing directly via pip3 install vocalinux instead of install.sh succeeded without issue. Since Mint is explicitly listed as a distribution the maintainer wants tested, it may be worth having install.sh check the Ubuntu/Debian version and skip or substitute this dependency on 22.04/Debian 11 bases.

flesler:
## Summary

- Remove `util-linux-extra` from the unconditional `APT_PACKAGES_UBUNTU` list
- Add an Ubuntu version gate (24.04+) matching the existing Debian 13+ pattern
- Add a regression test asserting the base Ubuntu package list no longer hard-requires `util-linux-extra`

## Problem

`install.sh` fails on Ubuntu 22.04 and derivatives (Linux Mint, Pop, etc.) with:

```
E: Unable to locate package util-linux-extra
```

`util-linux-extra` only exists from Ubuntu 24.04 / util-linux 2.38 onward. On 22.04 the tools ship in the main `util-linux` package, which is always installed.

Fixes #526

## Test plan

- [x] `python3 tests/test_installer_cuda_diagnostics.py InstallerCudaDiagnosticsTests.test_ubuntu_22_package_list_omits_util_linux_extra`
- [x] On Ubuntu 22.04: `./install.sh --auto` deps step completes without requesting `util-linux-extra` (verified with early-exit harness, stdin from `/dev/null`)


jatinkrmalik (changes requested):
Can you take a look at the bot feedback above?

flesler:
Tested locally on Ubuntu 22.04 (Casius) with the updated `apt-cache show` probe — `./install.sh --auto --skip-models --no-rebuild-whispercpp` completed successfully, deps step passed without requesting `util-linux-extra`, and the regression test passes. Looks good from here.

jatinkrmalik (dismissed):
Thanks for the quick follow-up, @flesler. The `apt-cache show` probe is exactly what we needed, and it handles Ubuntu derivatives cleanly. Appreciate the local 22.04 verification too.

LGTM 👍

jatinkrmalik:
@flesler I cannot merge this because it seems like there is a merge conflict. Can you please rebase it to latest main?

jatinkrmalik:
Nevermind, I was able to just push affix to the pull request. It looks good, we'll merge it now.

PR #684, opened 2026-08-16
fix(startup): resolve model size per engine instead of the legacy key

### LINKED ISSUE #681: Startup resolves model size from the legacy generic key instead of get_model_size_for_engine(), loading the wrong model when the keys diverge
## Summary

Process startup resolves the model size from the legacy generic `model_size` key instead of `ConfigManager.get_model_size_for_engine()`. The generic key always holds the *last saved engine's* model (that's what `set_model_size_for_engine()` writes for backward compatibility), so whenever the engine and the generic key diverge, startup silently initializes the engine with the wrong model — typically reporting a perfectly valid downloaded model as missing and offering to download a different one.

To be fair: the settings dialog itself keeps the keys in sync (every engine save goes through `set_model_size_for_engine()`), so you cannot produce the divergence with today's dialog alone. The point of this issue is that the invariant is implicit and nothing else honors or enforces it:

- `vocalinux --engine whisper_cpp` (no `--model`) takes the engine from the CLI but the model from the generic key — instant mismatch if the last save was for another engine.
- Configs written by older releases or edited by hand stay divergent forever; `_migrate_config()` only ever creates `vosk_model_size`/`whisper_model_size` and never `whisper_cpp_model_size`.
- Any future writer that changes the engine without re-saving a model (a tray quick-switch, or the one-click download proposed in #682) would reintroduce the bug invisibly.

Meanwhile the correct resolver already exists and is what the dialog uses — and it falls back to the generic key when per-engine keys are absent, so switching startup to it is strictly safer.

## Reproduction (v0.15.0, `c3a9695`)

`~/.config/vocalinux/config.json` (a state any of the paths above can produce; here `medium` was last saved for VOSK):

```json
"speech_recognition": {
    "engine": "whisper_cpp",
    "model_size": "medium",
    "whisper_cpp_model_size": "small",
    ...
}
```

with a fully working `ggml-small.bin` present in `~/.local/share/vocalinux/models/whispercpp/`. Startup log:

```
INFO - Using model=medium (from saved config)
INFO - Final settings: engine=whisper_cpp, language=en-us, model=medium
INFO - whisper.cpp model 'medium' not found at ~/.local/share/vocalinux/models/whispercpp/ggml-medium.bin. Will download when needed.
```

Any dictation attempt then plays the error sound and shows the "No Speech Model" notification, despite a valid model for the configured engine being on disk. From the user's perspective the model they downloaded has "disappeared".

## Fix

`src/vocalinux/main.py` (~line 400):

```python
model_size = saved_settings.get("model_size", args.model)
```

→

```python
model_size = config_manager.get_model_size_for_engine(engine)
```

(keeping the CLI `--model` override as-is), and treat the generic `model_size` as a write-only legacy key.

## Related hardening (model state)

While digging into this I noticed the "does the model exist" state is derived from bare `os.path.exists()` everywhere, which makes this class of bug easy to hit and hard to notice:

- The VOSK zip is downloaded directly into `models/` without a `.tmp` suffix, and a failed extraction leaves a partial model directory behind that then passes the existence check (`recognition_manager.py`, the commented-out cleanup around the `BadZipFile` handler).
- Nothing records that a model failed to load; the app happily retries the same broken path on every start.

Building on the model management UI from #671, it would help if the Models section showed the *effective* model per engine (what startup will actually load, not just what is selected) — that alone would have made this mismatch visible immediately.

kacperpaczos:
Fixes #681.

## Problem

Startup resolved the model from the generic `model_size` key:

```python
model_size = saved_settings.get("model_size", args.model)
```

`set_model_size_for_engine()` overwrites that key on every save for backward compatibility, so it always holds *whichever engine was saved last*. When it disagrees with the configured engine, startup initializes that engine with a foreign model — the engine's own `<engine>_model_size` value is ignored.

Observed on v0.15.0 with `engine: whisper_cpp`, `model_size: medium` (left by a VOSK save) and `whisper_cpp_model_size: small`, with a valid `ggml-small.bin` on disk:

```
INFO - Using model=medium (from saved config)
INFO - Final settings: engine=whisper_cpp, language=en-us, model=medium
INFO - whisper.cpp model 'medium' not found at .../whispercpp/ggml-medium.bin. Will download when needed.
```

Dictation then fails with the "No Speech Model" notification and offers a 1.5 GB download, while the model the engine is configured to use is present. The settings dialog keeps the keys in sync on save, so this is reached through `--engine` without `--model`, configs from older releases (`_migrate_config()` never creates `whisper_cpp_model_size`), and hand-edited configs.

## Change

1. Startup uses the resolver the settings dialog already uses, with `--model` keeping precedence:

   ```python
   model_size = config_manager.get_model_size_for_engine(engine)
   ```

2. `_migrate_config()` now also copies `model_size` into `whisper_cpp_model_size` when the user file lacks it (the follow-up from review). Without this, the resolver's generic-key fallback never fires for whisper.cpp — `load_config()` merges `DEFAULT_CONFIG`, which already carries `whisper_cpp_model_size: "tiny"` — so a pre-per-engine config holding only `model_size: small` would start on default `tiny` instead of `small`. `_check_needs_migration()` now also triggers on a missing `whisper_cpp_model_size`.

## Tests

- `test_migrate_copies_model_size_to_whisper_cpp` — real `ConfigManager` on a temp config holding only `{engine: whisper_cpp, model_size: small}` ends up on `small`. Red without the migration hunk.
- `test_migrate_leaves_other_engines_on_their_defaults` — the migration does not hand one engine's model to the others.
- `test_resolver_prefers_the_engine_key_over_the_generic_one` — real `ConfigManager`, no mocks: a stale generic `medium` with `whisper_cpp_model_size: small` resolves to `small`. Locks the fallback story from review.
- `test_model_resolved_per_engine_not_from_generic_key` — startup wiring: `SpeechRecognitionManager` receives the per-engine model, not the generic key. Red without the `main.py` hunk.
- Full suite diffed against `main` baseline: no new failures. flake8/black/isort as CI runs them: clean.

### DIFF
--- a/src/vocalinux/main.py
+++ b/src/vocalinux/main.py
@@ -397,8 +397,11 @@ def main():
         model_size = args.model
         logger.info(f"Using model={model_size} (from command line)")
     else:
-        model_size = saved_settings.get("model_size", args.model)
-        logger.info(f"Using model={model_size} (from saved config)")
+        # Resolve per engine: the generic "model_size" key always holds the
+        # model of whichever engine was saved last, so reading it directly
+        # loads the wrong model whenever the two disagree.
+        model_size = config_manager.get_model_size_for_engine(engine)
+        logger.info(f"Using model={model_size} (saved for engine {engine})")
 
     vad_sensitivity = saved_settings.get("vad_sensitivity", 3)
     silence_timeout = saved_settings.get("silence_timeout", 2.0)
--- a/src/vocalinux/ui/config_manager.py
+++ b/src/vocalinux/ui/config_manager.py
@@ -189,7 +189,9 @@ class ConfigManager:
         sr_config = user_config.get("speech_recognition", {})
         # Need migration if we have model_size but not the per-engine keys
         return "model_size" in sr_config and (
-            "vosk_model_size" not in sr_config or "whisper_model_size" not in sr_config
+            "vosk_model_size" not in sr_config
+            or "whisper_model_size" not in sr_config
+            or "whisper_cpp_model_size" not in sr_config
         )
 
     def _migrate_config(self, user_config: dict):
@@ -213,6 +215,18 @@ class ConfigManager:
                 current_model if current_engine == "whisper" else "tiny"
             )
             logger.info(f"Migrated whisper_model_size to: {sr_config['whisper_model_size']}")
+
+        if "whisper_cpp_model_size" not in user_sr_config:
+            # DEFAULT_CONFIG already carries whisper_cpp_model_size, so the
+            # generic fallback in get_model_size_for_engine() never fires for
+            # this engine. Without this copy a config that only stores
+            # model_size would silently start on the default instead.
+            sr_config["whisper_cpp_model_size"] = (
+                current_model if current_engine == "whisper_cpp" else "tiny"
+            )
+            logger.info(
+                f"Migrated whisper_cpp_model_size to: {sr_config['whisper_cpp_model_size']}"
+            )
 
         self.save_config()
         logger.info("Config migrated to new per-engine model format")
--- a/tests/test_config_manager.py
+++ b/tests/test_config_manager.py
@@ -281,6 +281,59 @@ class TestConfigManager(unittest.TestCase):
         self.assertEqual(config_manager.config["speech_recognition"]["vosk_model_size"], "medium")
         self.assertEqual(config_manager.config["speech_recognition"]["whisper_model_size"], "tiny")
 
+    def test_migrate_copies_model_size_to_whisper_cpp(self):
+        """A whisper.cpp config that predates the per-engine keys keeps its model.
+
+        DEFAULT_CONFIG already ships whisper_cpp_model_size, so the generic
+        fallback inside get_model_size_for_engine() can never fire for this
+        engine: after the merge the key is always present. Only the migration
+        can carry the user's own model_size across (#681).
+        """
+        with open(self.temp_config_file, "w") as f:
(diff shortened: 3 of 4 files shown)


jatinkrmalik (changes requested):
@kacperpaczos Thank you for raising this.

The call site is right. Once engine is resolved, model size has to go through `get_model_size_for_engine()`, and `--model` has to keep precedence. That matches the v0.15.0 log in #681.

One claim in the PR body is not true on a real `ConfigManager` load. `get_model_size_for_engine()` only falls back to generic `model_size` when `<engine>_model_size` is missing from the in-memory dict. `load_config()` starts from `DEFAULT_CONFIG`, which already has `whisper_cpp_model_size: "tiny"`, and `_migrate_config()` still never writes that key. A file that is only `{engine: whisper_cpp, model_size: small}` used to start on `small` and will now start on default `tiny`. That is the older-config path #681 already called out.

Not a reason to revert this hunk. The settings dialog already uses this resolver. Can we drop "older configs behave exactly as before" from the PR text?

Follow-up I am fine doing later: extend `_migrate_config` to copy `model_size` into `whisper_cpp_model_size` when the user file lacks it.

The new test would fail on current `main` because `get_model_size_for_engine` is never called. It does not run the real resolver (the method is mocked), so default-merge and missing-key fallback cannot fail this test. A second case with a real `ConfigManager` on a temp config, or `--engine whisper_cpp` without `--model`, would lock the fallback story.

jatinkrmalik (approved):
Follow-ups look good. The per-engine migrate is in, the tests hit a real ConfigManager, and CI is green.

PR #685, opened 2026-08-16
fix(settings): save the model only once the engine loaded it

kacperpaczos:
Addresses the settings-dialog half of #681 (see [this comment](https://github.com/VocaHQ/vocalinux/issues/681#issuecomment-5308738511)). Independent of #684, which fixes how startup resolves the model.

## 1. Settings were saved before the download that they depend on

`_apply_settings_internal()` persisted the selection and only then called `reconfigure()` — and `reconfigure()` is what downloads a missing model. Observed on v0.15.0, picking a 1.5 GB variant:

```
19:27:51,284 - config_manager - INFO - Set whisper_cpp model size to: medium.en
19:27:51,284 - config_manager - INFO - Saved configuration to ~/.config/vocalinux/config.json
19:27:51,647 - recognition_manager - INFO - Downloading whisper.cpp 'medium.en' model...
```

The config is committed 363 ms before the download starts and is never rolled back, so a download that fails, is cancelled, or is interrupted leaves the config naming a model that is not on disk. The next start logs `model 'medium.en' not found` and dictation shows "No Speech Model", while the model that worked five minutes ago is still on disk — just no longer selected. This is the mechanism behind user reports of a downloaded model "disappearing".

Now the engine is reconfigured first and the config is written only after it accepted the settings.

## 2. A failed apply left the pickers on the rejected model

That also blocked retrying: the combo still showed the failed model, and re-selecting the same entry emits no `changed` signal, so there was no way to trigger the download again from the UI. `_resync_model_ui_from_config()` puts the pickers back on what is actually saved, from both the modal download path and the plain apply path.

## 3. Selecting an engine applied nothing

`_on_engine_changed()` repopulated the pickers and the info card but never called `_auto_apply_settings()`, unlike every other control on the page. The dialog showed the new engine while the config and the running recognizer stayed on the old one, and closing the dialog discarded the change silently — the model card meanwhile read "Downloaded and ready ✓" for an engine that was not running. Observed live: engine switched to whisper.cpp in the UI, config still `engine: vosk` three minutes later.

Engine changes now apply like the rest. Remote API is the one exception: it is deferred until a server URL exists, because applying it empty can only fail validation, and `_on_remote_api_settings_changed()` already applies as soon as the URL is entered.

Note this means selecting an engine whose model is not downloaded now opens the existing download dialog (with its size and Cancel button) rather than doing nothing.

## Follow-up included (#692)

An earlier revision left one gap: after a failed or cancelled engine switch the engine picker kept displaying the engine that was never applied — filed as #692. The review was right that the resync sat in an `except` the download thread could not reach: `_apply_settings_internal()` catches internally and returns `False` unless `raise_errors=True` makes it raise. Both sides of that contract are now handled, without depending on any other PR:

- both download threads check the **return value**: `False` schedules `_resync_model_ui_from_config()` and reports `set_complete(False, ...)` instead of claiming success;
- the `except` handlers (cancel included) also schedule the resync;
- **backstop**: after `download_dialog.run()` returns, `_resync_engine_ui_if_unapplied()` compares the engine on screen with the engine in the config and resyncs on mismatch — whatever ended the modal, including paths not enumerated above;
- the same backstop runs on dialog close (`CLOSE` / `DELETE_EVENT`), which also covers the Remote API-without-URL deferral: the combo cannot outlive the config it never reached.

The resync itself holds the apply-suppression flag, so `_on_engine_changed()` repaints without cascading into another apply, and `_on_engine_changed()` now checks `_initializing`/`_applying_settings` itself like every other control on the page (a programmatic repaint no longer rewrites the language either).

## Tests

The earlier revision's `_load_settings_dialog()` ran at collection time and restored only `sys.modules`, not the package attribute — the cause of the 3.9/3.10 `test_model_deletion` failures. The reimport now lives in a **module-scoped fixture** that restores **both** `sys.modules` and `vocalinux.ui.settings_dialog` on the package, so collection touches no interpreter state.

`tests/test_settings_model_state.py` now reaches the modal path the review asked for (download thread run inline, `GLib.idle_add` captured):

- `needs_download=True` and `_apply_settings_internal` **returns `False`** (not raises): resync is scheduled, `set_complete(True, "")` is not, nothing is saved;
- cancelled download: resync scheduled, `set_complete(False, "Download cancelled")`;
- modal close and dialog close each trigger `_resync_engine_ui_if_unapplied()`; the mismatch check resyncs and the matching case leaves the UI alone;
- save-after-reconfigure ordering, no write when reconfiguring fails, engine change applies, Remote API deferred until a URL exists, and a programmatic repaint neither applies nor rewrites the language.

`_dialog_stub` now sets `speech_engine.state = RecognitionState.IDLE` (the enum, not a string), so no test sleeps in the `was_running` branch.

Each behavioural test was verified red with its hunk reverted. Full suite diffed against the `main` baseline: no new failures. flake8/black/isort as CI runs them: clean.

### DIFF
--- a/src/vocalinux/ui/settings_dialog.py
+++ b/src/vocalinux/ui/settings_dialog.py
@@ -3850,6 +3850,7 @@ class SettingsDialog(Gtk.Dialog):
         """Persist deferred text edits before the settings dialog closes."""
         if response_id in (Gtk.ResponseType.CLOSE, Gtk.ResponseType.DELETE_EVENT):
             self._flush_advanced_prompt_if_dirty()
+            self._resync_engine_ui_if_unapplied()
 
     def _on_advanced_prompt_changed(self, buffer):
         """Track prompt edits without applying settings on every keystroke."""
@@ -4476,7 +4477,14 @@ class SettingsDialog(Gtk.Dialog):
 
         engine = _engine_from_display(engine_text)
 
-        current_lang = self.language_combo.get_active_id()
+        # A programmatic change — the resync after a failed apply — must repaint
+        # for the engine now shown, but must not rewrite the language the user
+        # picked, and must not cascade into another apply. Every other control on
+        # this page checks these flags itself rather than leaning on
+        # _auto_apply_settings to do it.
+        programmatic = self._initializing or self._applying_settings
+
+        current_lang = None if self._applying_settings else self.language_combo.get_active_id()
         if current_lang:
             if engine == "vosk" and (
                 current_lang == "auto" or not SUPPORTED_LANGUAGES.get(current_lang, {}).get("vosk")
@@ -4490,6 +4498,9 @@ class SettingsDialog(Gtk.Dialog):
         self._update_engine_specific_ui()
         self._update_model_info()
         self._update_voice_commands_for_engine()
+
+        if programmatic:
+            return
 
         # Every other control on this page applies as soon as it changes. Without
         # this the new engine is only displayed: the config and the running
@@ -4855,9 +4866,21 @@ class SettingsDialog(Gtk.Dialog):
                         cancel_check_id = GLib.timeout_add(100, check_cancelled)
 
                         try:
-                            self._apply_settings_internal(settings, raise_errors=True)
-                            GLib.idle_add(download_dialog.set_complete, True, "")
-                            GLib.idle_add(self._populate_model_options)
+                            applied = self._apply_settings_internal(settings, raise_errors=True)
+                            if applied:
+                                GLib.idle_add(download_dialog.set_complete, True, "")
+                                GLib.idle_add(self._populate_model_options)
+                            else:
+                                # raise_errors makes failures raise today, but a
+                                # False return must not be read as success: it
+                                # means nothing was saved and the pickers still
+                                # show settings the engine never took.
+                                GLib.idle_add(self._resync_model_ui_from_config)
+                                GLib.idle_add(
+                                    download_dialog.set_complete,
+                                    False,
+                                    "Could not apply the new settings",
+                                )
                         finally:
                             GLib.source_remove(cancel_check_id)
                             self.speech_engine.set_download_progress_callback(None)
@@ -4887,6 +4910,9 @@ class SettingsDialog(Gtk.Dialog):
                 threading.Thread(target=download_and_apply, daemon=True).start()
                 download_dialog.run()
                 download_dialog.destroy()
+                # Whatever ended the modal — cancel, failure, or a path not
+                # covered above — the pickers must not outlive the config.
+                self._resync_engine_ui_if_unapplied()
                 return
 
             logger.info(f"Auto-applying settings: {settings}")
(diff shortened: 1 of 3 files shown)


jatinkrmalik (changes requested):
@kacperpaczos Thank you for raising this. Save-after-`reconfigure` is the right fix for the disappearing-model write. I still cannot merge this.

**CI**

3.9 and 3.10 fail on an untouched test:

```
tests/test_model_deletion.py::test_whisper_list_and_delete
AssertionError: assert [] == ['tiny']
```

That is on both `7151b69` and `0c7ba5d`. `main` CI is green. The only new import-time side effect is `tests/test_settings_model_state.py` calling `_load_settings_dialog()` at collection. It pops `vocalinux.ui.settings_dialog` from `sys.modules`, reimports it under a patched `Gtk`, then restores `saved` in `sys.modules` but never binds `vocalinux.ui.settings_dialog` on the package.

After collection, `sys.modules` and `vocalinux.ui.settings_dialog` can be different module objects. `from vocalinux.ui.settings_dialog import _list_downloaded_whisper_models` and `patch("vocalinux.ui.settings_dialog._get_whisper_cache_dir")` then disagree. The list helper sees the real cache, `expanduser` is pointed at `missing_default`, result is `[]`.

#690 uses this same helper in `test_download_error_reporting.py` (collects *before* `test_model_deletion`) and 3.9/3.10 pass there. After restore, please also set the package attribute to `saved`, or keep one module object in `sys.modules` and do not pop it. I usually skip failing pipelines, so can you get 3.9/3.10 green first?

**#692 is not actually closed**

The resync in the download-thread `except` is unreachable for a failed or cancelled download. `_apply_settings_internal()` still catches every exception and returns `False`. This thread therefore always falls through to `set_complete(True, "")` and `_populate_model_options()` for the *displayed* engine, then returns after `download_dialog.run()` with no resync.

Repro from #692: switch to an engine whose model is not on disk, cancel the modal. Config is correctly not written. The engine combo still shows the engine that was never applied.

Please do not depend on #690 to make this `except` live. Check the return value, or resync after the modal closes if `get_selected_settings()` does not match the saved engine:

```python
ok = self._apply_settings_internal(settings)
if ok:
    GLib.idle_add(download_dialog.set_complete, True, "")
    GLib.idle_add(self._populate_model_options)
else:
    GLib.idle_add(self._resync_model_ui_from_config)
    GLib.idle_add(download_dialog.set_complete, False, "...")
```

`apply_settings()`'s modal `except` never calls resync at all.

**Tests**

None of the new tests hit the modal path that #692 is about.

- `test_failed_download_*` only calls `_apply_settings_internal` (and the name is wrong)
- `test_failed_auto_apply_*` patches `_is_vosk_model_downloaded` True, so `_auto_apply_settings` never opens `ModelDownloadDialog`

Please add a test where `needs_download` is True, `_apply_settings_internal` returns `False` (not raises), and assert `_resync_model_ui_from_config` is scheduled / the engine combo is restored. Today that test would fail on this branch.

Also: `_dialog_stub` sets `speech_engine.state = "idle"` (str). That is not `RecognitionState.IDLE`, so the first two tests sleep 0.5s on the `was_running` branch.

**Nit**

Remote API with an empty URL still leaves the combo on an engine that was never applied. Same UI/config split, documented as intentional. If that stay-deferred is on purpose, resync the combo on dialog close when the saved engine differs, or disable the Remote API row until a URL exists.

`_on_engine_changed` also has no `_initializing` / `_applying_settings` guard (every other control on this page does). Apply is only skipped because `_auto_apply_settings` checks the flags; the language rewrite and repaint still run during resync.

Happy to re-review once 3.9/3.10 are green and the cancel-download path actually resyncs.

jatinkrmalik (approved):
Follow-ups look good. Cancel and failed download now resync the engine picker, the tests hit that modal path, and CI is green.

PR #686, opened 2026-08-16
fix(settings): size the unused downloads list to its real rows

### LINKED ISSUE #683: Unused downloads list clips its rows — header says 2 leftover models, only 1 is visible
## Symptom

v0.15.0, Speech Model page, whisper.cpp engine, three models on disk (`ggml-tiny.bin`, `ggml-small.bin`, `ggml-medium.en.bin`), active model `medium.en`.

The **Unused downloads** expander header correctly says **"2 leftover models"**, but the list below renders only **one** row (Tiny). The second unused model (Small, 466 MB) is nowhere to be seen and cannot be deleted — no scrollbar, no visual hint that the list is clipped.

Before the third model was downloaded the section behaved correctly ("1 leftover model" → Tiny listed), so it is the row *added by the refresh* that ends up invisible.

## Diagnosis

The header and the row loop use the same list (`_refresh_unused_downloads()` in `settings_dialog.py`), so both rows are definitely created — the missing entry is a rendering problem, not a data problem:

```python
self.unused_models_scroll.set_min_content_height(min(220, 56 * count))
```

- A `PreferenceRow` is not 56 px tall: 12 px top + 12 px bottom margin plus a title+subtitle stack / a Delete button makes it ~64–70 px. Two rows ≈ 130 px get a 112 px viewport.
- The scroll relies on `set_propagate_natural_height(True)` through a hand-built `Box`-forced `Viewport` (see the comment at the `unused_models_scroll` setup about making "the expander child report a real height") — when that propagation doesn't kick in, the viewport stays at the underestimated `min_content_height`, and with `PolicyType.AUTOMATIC` the overlay scrollbar is invisible until hovered, so the clipped tail of the list gives no affordance at all.
- The refresh also restores the previous expanded state after rebuilding rows; when the rebuild happens while the expander is collapsed, the stale allocation makes the clipping worse on the next expand.

## Suggested fix

Don't guess the row height. Either drop the manual `min_content_height` math and let the natural height propagate (cap only with `set_max_content_height(220)`), or compute the height from the actual row's `get_preferred_height()`. If the list can legitimately overflow the 220 px cap, switch the policy so the scrollbar is visible whenever content is clipped.

kacperpaczos:
Fixes #683.

## Problem

`_refresh_unused_downloads()` sized the list from a per-row estimate, before the rows existed:

```python
self.unused_models_scroll.set_min_content_height(min(220, 56 * count))
```

A `PreferenceRow` does not render in 56px — 12px top and bottom margins around a title/subtitle stack next to a Delete button make it noticeably taller. With two leftover whisper.cpp models the expander header correctly said "2 leftover models" while the viewport only had room for the first row: the second was clipped below the edge, so it could neither be seen nor deleted. Nothing signalled the clipping either, because `AUTOMATIC` policy draws an overlay scrollbar that stays invisible until hovered.

The header and the rows come from the same list, so the count was always right — only the height was guessed.

## Change

- Measure the list after its rows are in place (`get_preferred_height()`) and cap that at the existing 220px, now a named constant. The clamp itself is a module-level helper, `_clamp_unused_downloads_height()`.
- Turn off overlay scrolling for this list, so when a genuinely long list does hit the cap the scrollbar takes part in the layout and the clipping is visible.
- Backstop for the collapsed-measure question from review: the refresh measures while the expander is collapsed and the dialog may not be mapped yet. `notify::expanded` now remeasures, so even if that first `get_preferred_height()` ever came back short, expanding the list fixes the height before the user can see clipped rows.

## Tests

The previous revision reimported `vocalinux.ui.settings_dialog` at collection time and restored only `sys.modules`, which is what broke `test_model_deletion` on 3.9/3.10. That is gone: `tests/test_unused_downloads_height.py` now touches no interpreter state at all. The arithmetic lives in the pure helper and is tested directly; the wiring is asserted with `inspect.getsource`, the existing idiom from `test_settings_mode_change.py`:

- a list whose rows render taller than the old 56px estimate gets its full measured height, and a very long list is capped instead of pushing the page down;
- `_refresh_unused_downloads` calls `_fit_unused_downloads_height()` and the per-row estimate does not come back;
- `_fit_unused_downloads_height` measures the listbox and feeds the scroll;
- overlay scrolling stays off; expanding the expander remeasures.

Each wiring test goes red when its hunk is reverted. Full suite diffed against the `main` baseline: no new failures. flake8/black/isort as CI runs them: clean.

### DIFF
--- a/src/vocalinux/ui/settings_dialog.py
+++ b/src/vocalinux/ui/settings_dialog.py
@@ -2438,6 +2438,12 @@ class SettingsDialog(Gtk.Dialog):
         list_holder.pack_start(self.unused_models_group.listbox, False, False, 0)
         self.unused_models_scroll.add(list_holder)
         self.unused_expander.add(self.unused_models_scroll)
+        # Backstop: the refresh measures while the expander is collapsed and
+        # the dialog may not be mapped yet. Remeasure when the user expands,
+        # so a short first measurement can never leave the list clipped.
+        self.unused_expander.connect(
+            "notify::expanded", lambda *_args: self._fit_unused_downloads_height()
+        )
         self.unused_models_group.pack_start(self.unused_expander, False, False, 0)
         self.content_box.pack_start(self.unused_models_group, False, False, 0)
 
--- a/src/vocalinux/ui/settings_dialog.py
+++ b/src/vocalinux/ui/settings_dialog.py
@@ -143,6 +143,9 @@ ENGINE_MODELS = {
     "remote_api": [],  # Remote API does not need local models
 }
 
+# Tallest the "Unused downloads" list may grow before it starts scrolling
+_UNUSED_DOWNLOADS_MAX_HEIGHT = 220
+
 # Engine display name mapping
 ENGINE_DISPLAY_NAMES = {
     "vosk": "Vosk",
@@ -2417,7 +2420,10 @@ class SettingsDialog(Gtk.Dialog):
         self.unused_models_scroll.set_policy(Gtk.PolicyType.NEVER, Gtk.PolicyType.AUTOMATIC)
         self.unused_models_scroll.set_shadow_type(Gtk.ShadowType.NONE)
         self.unused_models_scroll.set_propagate_natural_height(True)
-        self.unused_models_scroll.set_max_content_height(220)
+        self.unused_models_scroll.set_max_content_height(_UNUSED_DOWNLOADS_MAX_HEIGHT)
+        # Keep the scrollbar in the layout: an overlay one stays invisible until
+        # hovered, so a list clipped by the cap looks like it has no more rows.
+        self.unused_models_scroll.set_overlay_scrolling(False)
 
         # ListBox is GtkScrollable, so wrapping it in a Box forces a Viewport
         # and lets the expander child report a real height.
@@ -4386,7 +4392,6 @@ class SettingsDialog(Gtk.Dialog):
         count = len(unused)
         leftover = "leftover model" if count == 1 else "leftover models"
         self.unused_expander.set_label(f"Unused downloads ({count} {leftover})")
-        self.unused_models_scroll.set_min_content_height(min(220, 56 * count))
 
         was_expanded = self.unused_expander.get_expanded()
         for model_id, title, size_label in unused:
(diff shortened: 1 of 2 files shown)


jatinkrmalik (changes requested):
@kacperpaczos Thank you for raising this. The GTK change itself looks right, but I cannot merge while 3.9/3.10 are red.

The new helper is why those jobs failed. `_load_settings_dialog()` runs at import time, pops `vocalinux.ui.settings_dialog` from `sys.modules`, reimports it under a patched `Gtk`, then restores only `sys.modules`, not `vocalinux.ui.settings_dialog` on the package. `test_settings_mode_change.py` already imported that module during collection.

Result on this PR, not on `main`: `tests/test_model_deletion.py::test_whisper_list_and_delete` fails on 3.9 and 3.10 with `assert [] == ['tiny']`. That is what happens when `_list_downloaded_whisper_models` and `patch("vocalinux.ui.settings_dialog._get_whisper_cache_dir")` are not the same module object. Unpatched `_get_whisper_cache_dir()` uses `MODELS_DIR`, and the test's `os.path.expanduser` patch points the fallback at a missing directory.

p.s. I usually skip failing pipelines and assume the author will make them review-ready. Can you extract the `min()` assert into a pure helper and drop the reimport, or restore the package attribute and import inside a fixture? Please do not touch `sys.modules` at collection time.

The measure-then-cap change is the right GTK3 fix for #683. `220` is only the existing viewport cap, not a per-row guess. Neither test calls `_refresh_unused_downloads`, builds a `PreferenceRow`, or checks `set_overlay_scrolling(False)`. They would pass if `_fit_unused_downloads_height` were never called from refresh.

One more thing to confirm: `_fit_unused_downloads_height()` runs after `set_expanded(was_expanded)`. Default is collapsed, and the first refresh is during dialog init. Does `get_preferred_height()` stay non-zero when the expander is collapsed and the dialog is not yet mapped? If that comes back short, remeasure on map or expander expand.

Happy to re-review once 3.9/3.10 are green.

jatinkrmalik (approved):
Follow-ups look good. The height clamp is a pure helper, the collection-time reimport is gone, it remeasures on expand, and CI is green including 3.9/3.10.

PR #687, opened 2026-08-16
feat(tray): offer the recommended model from the missing-model notification

### LINKED ISSUE #682: Add a one-click download action to the 'No Speech Model' notification
## Current behavior

When dictation is triggered and no model is available for the configured engine, the app plays the error sound and shows a plain desktop notification (v0.15.0, `recognition_manager.py`, `_show_notification` → `notify-send`):

> **No Speech Model**
> Please open Settings and download a speech recognition model to use dictation.

The notification has no actions. The user has to open Settings, navigate to the Speech section, find the right model in the picker and start the download manually — several steps at exactly the moment they just wanted to dictate something.

## Proposal: one-click recovery

Add a default action button to that notification that downloads the **recommended** model for the active engine, reusing the recommendation logic that already exists (the same one that marks a model with ★ in the settings UI):

- whisper.cpp: `get_recommended_model()` in `utils/whispercpp_model_info.py` (hardware-aware: Vulkan/CUDA/VRAM/RAM) and its language-aware wrapper `_get_recommended_whispercpp_model_for_language()` in the settings dialog;
- Whisper / VOSK: `_get_recommended_whisper_model()` / `_get_recommended_vosk_model()` in `settings_dialog.py`.

So the button reads e.g. **"Download small (466 MB)"** — where `small` is whatever the existing recommendation logic picks for this machine and language, with the size shown so the user knows what one click commits them to.

Implementation notes:

- Use libnotify actions (or `notify-send -A` / GNotification) so the button appears on GNOME/KDE; on notification servers without action support, fall back to the current text-only notification (or make clicking the notification open Settings on the Speech page).
- Reuse the existing download plumbing (`_download_whispercpp_model` / `_download_vosk_model` and the progress callback) — show progress either via the existing `ModelDownloadDialog` or replace/update the notification with a completion message.
- When the download finishes, save the choice via `set_model_size_for_engine()`, load the model and (optionally) start listening again so the original intent — "I pressed the shortcut to dictate" — completes without further interaction.

## Note

This pairs with #681: model-state resolution must be per-engine, otherwise the one-click action could offer a model that doesn't match what the engine will actually load.

kacperpaczos:
Fixes #682.

## Problem

Pressing the dictation shortcut with no model on disk produced a dead end:

> **No Speech Model**
> Please open Settings and download a speech recognition model to use dictation.

No action, no button — and there is no explicit Download control in Settings either, since downloads only start as a side effect of changing a selection. The user has to open Settings, find the right entry in the model picker and change it, at exactly the moment they wanted to talk.

## Change

The notification now carries a **Download <model> (<size>)** button for the model this machine should use — the same one the picker marks with a star:

- whisper.cpp: `get_recommended_model()` (Vulkan/CUDA/VRAM/RAM), narrowed to the English-only variant when an English language is selected, via the existing `_recommended_whispercpp_variant_for_language()`;
- VOSK and Whisper: their existing RAM-based recommendations.

Those three were already in `settings_dialog`; they are now reachable through one `recommended_model_for_engine(engine, language)` that returns the model id, the reason, and display/size labels, so the dialog and the notification cannot drift apart.

Clicking the button downloads in the background, refreshes the notification with progress every few seconds, and — only once the engine has loaded the model — saves it as that engine's model via `set_model_size_for_engine()`. A failed download saves nothing and says so.

Three review findings are folded in:

- **The recommendation is often exactly what is already configured** (first run: `whisper_cpp` + `tiny`, which is also what a CPU-only machine is recommended), and `reconfigure()` only re-initializes when engine, model or language change — so the old code was a no-op that then claimed success. `reconfigure()` now takes `force_reinit`, the tray passes it, and success ("Speech model ready", plus the config write) is only reported when `speech_engine.model_ready` is true afterwards; otherwise the failure notification shows.
- **The first "starting..." notification is created on the main loop**: the action callback is dispatched there by libnotify, the worker thread only receives the handle and routes every later `update`/`close`/`notify` through `_idle_once`. libnotify/GObject never gets touched off the main loop.
- **Servers without the `actions` capability get Settings-only copy** — no instructions to click a button that was never shown.

Deliberately not done: dictation is not auto-started afterwards. A 1.5 GB download takes minutes, and grabbing the microphone unannounced would surprise someone who has moved on. The success notification names the shortcut instead.

## Layering

`recognition_manager` gained a `set_model_missing_handler()` hook and calls it where it used to notify. The tray registers the handler, since it owns the config, the GTK main loop and the settings UI. Without a handler — or when the handler declines, e.g. for Remote API, which has nothing to download — the previous notification is shown unchanged, so nothing regresses for headless or partial setups.

`ui/notifications.py` is new: libnotify when the server advertises the `actions` capability (which also allows updating a notification in place for progress), falling back to spawning `notify-send` exactly as before. The offer is handed to the main loop with `idle_add`, because dictation can be triggered from the shortcut thread.

## Tests

`tests/test_model_missing_offer.py` covers the recommendation per engine (including the English-only variant and Remote API having nothing to offer), the button label and what it downloads, the fallback when the server has no action support, that a successful download is persisted for that engine and a failed one is not, that a second click cannot start a second download, that callbacks scheduled on the main loop return falsy so `GLib.idle_add` cannot re-run them, and that the engine only shows its own notification when nobody handled it.

The same-model no-op the review called out is covered from both ends:

- `test_the_model_already_configured_is_still_fetched` — the tray's download against an engine stub that gates re-init exactly the way `reconfigure()` does; red without `force_reinit=True` at the call site;
- `test_force_reinit_reinitializes_an_engine_that_did_not_change` / `test_an_unchanged_reconfigure_still_does_nothing_by_default` — the real `reconfigure()` method: `force_reinit` re-initializes with `_defer_download` off, and callers that only tweak VAD still pay nothing;
- `test_a_model_that_did_not_load_is_not_reported_as_ready` — no config write and no success toast when `model_ready` stays false;
- `test_the_first_notification_is_created_on_the_main_loop`.

Each of these was verified red with its hunk reverted. `black`/`isort`/`flake8` run exactly as CI does: clean (the previous revision's `black` failure is gone). Full suite diffed against the `main` baseline: no new failures.

### DIFF
--- a/src/vocalinux/speech_recognition/recognition_manager.py
+++ b/src/vocalinux/speech_recognition/recognition_manager.py
@@ -974,6 +974,8 @@ class SpeechRecognitionManager:
 
         # Download progress tracking
         self._download_progress_callback: Optional[Callable[[float, float, str], None]] = None
+        # Set by the UI layer to offer the download when a model is missing
+        self._model_missing_handler: Optional[Callable[[str], bool]] = None
         self._download_cancelled = False
         self._defer_download = defer_download
         self._model_initialized = False
@@ -2184,6 +2186,30 @@ class SpeechRecognitionManager:
         """
         self._download_progress_callback = callback
 
+    def set_model_missing_handler(self, handler: Optional[Callable[[str], bool]]):
+        """
+        Set a handler for dictation attempted without a model on disk.
+
+        The UI layer registers one to offer the download directly; without a
+        handler, or when it declines, a plain notification is shown instead.
+
+        Args:
+            handler: Function(engine) -> True if it took care of the situation,
+                     or None to clear
+        """
+        self._model_missing_handler = handler
+
+    def _notify_model_missing(self) -> bool:
+        """Let the registered handler take over the missing model, if it can."""
+        handler = self._model_missing_handler
+        if handler is None:
+            return False
+        try:
+            return bool(handler(self.engine))
+        except Exception as e:
+            logger.error(f"Model-missing handler failed: {e}", exc_info=True)
+            return False
+
     def cancel_download(self):
         """Request cancellation of the current download."""
         self._download_cancelled = True
@@ -2468,12 +2494,13 @@ class SpeechRecognitionManager:
                     "Please download via Settings."
                 )
                 play_error_sound()
-                _show_notification(
-                    "No Speech Model",
-                    "Please open Settings and download a speech recognition model "
-                    "to use dictation.",
-                    "dialog-warning",
-                )
+                if not self._notify_model_missing():
+                    _show_notification(
+                        "No Speech Model",
+                        "Please open Settings and download a speech recognition model "
+                        "to use dictation.",
+                        "dialog-warning",
+                    )
                 return False
 
         logger.info("Starting speech recognition")
--- a/src/vocalinux/speech_recognition/recognition_manager.py
+++ b/src/vocalinux/speech_recognition/recognition_manager.py
@@ -977,6 +977,10 @@ class SpeechRecognitionManager:
         # Set by the UI layer to offer the download when a model is missing
         self._model_missing_handler: Optional[Callable[[str], bool]] = None
         self._download_cancelled = False
+        # Held by whichever UI is downloading a model. The tray and the settings
+        # dialog both download in the background, and there is only one progress
+        # callback and one engine configuration between them.
+        self._download_claim = threading.Lock()
         self._defer_download = defer_download
         self._model_initialized = False
         # True while auto-pause has unloaded the model for a configured app/game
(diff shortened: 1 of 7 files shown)


jatinkrmalik (changes requested):
@kacperpaczos Thank you for raising this. I am going to request changes.

This does not download when the recommended id is already `self.speech_engine.model_size`. `reconfigure()` only re-inits when engine, model, or language change (`restart_needed`). `force_download` is ignored otherwise. It just logs `Applied VAD/silence timeout changes.`

First-run defaults are `whisper_cpp` + `tiny` + `language=auto`. On CPU-only under 16GB (and the no-psutil fallback) `get_recommended_model()` is also `tiny`, and `_language_is_english("auto")` is false, so this is a no-op. Same if the user already selected that model in Settings.

You then `set_model_size_for_engine`, `save_config`, and show "Speech model ready" because `reconfigure` did not raise. The next shortcut is the same missing-model notification.

Can we always force a download/re-init? For example `_defer_download = False` plus `reinitialize_after_resume()`, or a `force_reinit` on `reconfigure`. Persist and notify success only if `model_ready` is true afterwards. Please add a test where configured `model_size` already equals the recommendation.

`notifications.notify(...)` for the "starting..." toast runs on the download thread. Later `update` / `close` / `notify` correctly go through `_idle_once`. libnotify/GObject is not thread-safe. Please create that first notification on the main loop too.

The body always says "Download it from this notification" even when `supports_actions()` is false and `action` is `None`. Fallback servers get instructions to click a button that was never shown. Use Settings-only copy when there is no action, like the old notification.

Also, `black --check` fails on `tray_indicator.py` (blank line after `_idle_once`) and `tests/test_model_missing_offer.py` (wrapping). That skips Test Python. I usually skip failing pipelines, so can you run this and get CI green?

```bash
black src/ tests/
```

Scope is large, but it is not drive-by. The libnotify helper and shared `recommended_model_for_engine` are the right extractions. The tests just miss the same-size no-op, which is the case that breaks the feature.

jatinkrmalik (dismissed):
Follow-ups look good. Same-size recommended models actually download now, the first toast is on the main loop, fallback copy no longer mentions a missing button, and CI is green.

PR #691, opened 2026-08-16
fix(config): share one ConfigManager across the process

### LINKED ISSUE #689: Independent ConfigManager instances silently revert each other's saved settings
## Symptom

Saved settings can silently revert to earlier values. The mechanism needs no race and no threads — it is deterministic.

## Diagnosis

The process holds several independent `ConfigManager` instances:

- `main.py:345` — used for the first-run/autostart flow,
- `tray_indicator.py:107` — the tray's own instance, shared with the settings dialog,
- `ui/audio_feedback.py:59` — a fresh instance per call (read-only today).

Each instance loads the **whole** config into memory at construction time, and `save_config()` writes that **whole** in-memory cache back to disk. There is no reload before save, no merge, no lock. Any save from an instance whose cache is stale reverts everything another instance has saved in the meantime.

Deterministic reproduction (v0.15.0, `c3a9695`) — two instances, sequential calls, single thread:

```python
instance_a = ConfigManager()   # e.g. the tray's
instance_b = ConfigManager()   # e.g. main.py's

instance_a.update_speech_recognition_settings(
    {"engine": "whisper_cpp", "model_size": "small"}
)
instance_a.save_config()       # user's choice is on disk

instance_b.set("general", "autostart", True)
instance_b.save_config()       # unrelated toggle...
```

Result:

```
user saved engine        : whisper_cpp / small (via instance A)
engine on disk afterwards: vosk / medium
```

The unrelated autostart save undid the engine switch, because instance B wrote its stale copy of the `speech_recognition` section along with it.

## Impact

Today the in-process exposure is narrow — `main.py`'s instance only saves during the first-run flow, and the audio-feedback instances never save — so users mostly do not hit it. But it is a landmine: every future `ConfigManager()` caller that ever saves becomes a potential silent settings-wiper, the failure is invisible when it happens, and "my settings randomly reverted" is exactly the bug report it produces. The download thread paths that reconfigure and save make the window easier to hit over time.

## Suggested direction

Either a process-wide singleton (one shared instance, as the tray/dialog already share theirs), or make `save_config()` merge onto a fresh read of the file instead of writing the whole cache. A singleton is the smaller change and matches how the config is actually used.

kacperpaczos:
Fixes #689. Based on `main`, independent of the other open PRs.

## Problem

The process holds several independent `ConfigManager` instances — `main.py`, the tray (shared with the settings dialog), and a fresh one per sound-effects check. Each loads the whole config into memory at construction time and `save_config()` writes that whole cache back, so a save from a stale instance silently reverts everything another instance saved in the meantime. No race needed; the deterministic reproduction is in #689.

The codebase already works around this: `_should_append_trailing_space()` in `main.py` re-reads `config.json` from disk on every call, with a comment explaining that "TrayIndicator and main() each construct their own ConfigManager, so an in-memory read from main's instance would miss Settings writes".

## Change

`get_shared_config_manager()` in `config_manager.py` returns one process-wide instance; every production call site goes through it — `main.py`, the tray, and both sound-effects lookups in `audio_feedback.py` (the tone lookup arrived with #707 and constructed its own instance again; the source guard below caught it on rebase). Creation is behind a `threading.Lock`: `__init__` does `makedirs()` plus a full `load_config()`, so two threads racing the first call would otherwise each build an instance and reintroduce the overwrite. `ConfigManager` stays directly constructible (the config-manager unit tests rely on that), and the tray/dialog pair already shared one instance — this extends that arrangement to the rest of the process. The now-outdated comments on the disk-read workaround are updated; the workaround itself is left in place as the conservative path.

## Tests

- `tests/test_shared_config_manager.py`:
  - the accessor returns one instance (`first is second`, asserted before any mutation);
  - the two-writer scenario from #689 run through the accessor: **both handles are taken before either save**, then each saves — both writes survive. With two instances the second save would win with its stale cache;
  - `test_two_instances_still_revert_each_other`: characterization of the footgun — two direct `ConfigManager()` instances created before either save still lose the first write. This is the red-on-`main` shape of #689;
  - a source guard walks `src/` and fails if any module besides `config_manager.py` constructs `ConfigManager()` directly again. It fails on current `main` (three offender files) and passes with this change.
- `conftest.py` drops the shared instance between tests — in production the long-lived instance is the point, but in tests it would be created under whichever mocks the first caller had active and leak them into every later test.
- Tray, `test_main*` and `test_audio_feedback` tests patch `get_shared_config_manager` instead of the class, so a leftover singleton can never silently skip the mock.

### A test that only passed by accident

Two tests (`test_main_single_instance_already_running`, `test_main_notification_on_second_instance`) mocked the lock with `patch.dict(sys.modules, {"vocalinux.single_instance": ...})`. `main()` does `from . import single_instance`, which binds through the package attribute, so once any earlier test has imported the real module the mock is ignored and the real lock is taken. The tests still passed — but only because a `ConfigManager` mock from an *earlier* test had leaked into the tray module at import time and made startup fail with `sys.exit(1)` for an unrelated reason. This change removes that leak (the tray now resolves the config at call time), which exposed the broken mocking; both tests now patch `vocalinux.single_instance.acquire_lock` directly, which is order-independent.

Full suite matches the pre-existing baseline exactly; flake8 reports no new findings on any touched file.

### DIFF
--- a/src/vocalinux/main.py
+++ b/src/vocalinux/main.py
@@ -29,9 +29,10 @@ def _should_append_trailing_space() -> bool:
     """Return whether completed transcriptions should get a trailing space.
 
     Reads config.json from disk on each call so Settings toggles take effect
-    immediately. TrayIndicator and main() each construct their own
-    ConfigManager, so an in-memory read from main's instance would miss
-    Settings writes (same pattern as TextInjector._should_copy_to_clipboard).
+    immediately. Historically TrayIndicator and main() each constructed their
+    own ConfigManager, so an in-memory read would miss Settings writes; the
+    instance is shared now, but the disk read stays as the conservative path
+    (same pattern as TextInjector._should_copy_to_clipboard).
     """
     try:
         import json
@@ -325,7 +326,7 @@ def main():
     from .text_injection import text_injector
     from .ui import tray_indicator
     from .ui.action_handler import ActionHandler
-    from .ui.config_manager import ConfigManager
+    from .ui.config_manager import get_shared_config_manager
     from .ui.logging_manager import initialize_logging
 
     # Initialize logging manager early
@@ -342,7 +343,7 @@ def main():
     except Exception as e:
         logger.debug(f"Could not start IBus daemon: {e}")
 
-    config_manager = ConfigManager()
+    config_manager = get_shared_config_manager()
     saved_settings = config_manager.get_settings().get("speech_recognition", {})
     audio_settings = config_manager.get_settings().get("audio", {})
 
@@ -498,8 +499,7 @@ def main():
 
                 text_to_inject = capitalize_sentences(text_to_inject)
 
-            # Read from disk so the Settings toggle applies without restart
-            # (main and tray each own a ConfigManager instance).
+            # Read from disk so the Settings toggle applies without restart.
             append_trailing_space = _should_append_trailing_space()
 
             if append_trailing_space:
--- a/src/vocalinux/ui/audio_feedback.py
+++ b/src/vocalinux/ui/audio_feedback.py
@@ -59,9 +59,9 @@ ERROR_SOUND = _resource_manager.get_sound_path("error")
 
 def _is_sound_effects_enabled() -> bool:
     try:
-        from .config_manager import ConfigManager
+        from .config_manager import get_shared_config_manager
 
-        return ConfigManager().is_sound_effects_enabled()
+        return get_shared_config_manager().is_sound_effects_enabled()
     except Exception:
         return True
 
--- a/src/vocalinux/ui/audio_feedback.py
+++ b/src/vocalinux/ui/audio_feedback.py
@@ -68,9 +68,9 @@ def _is_sound_effects_enabled() -> bool:
 
 def _resolved_tone() -> str:
     try:
-        from .config_manager import ConfigManager
+        from .config_manager import get_shared_config_manager
 
-        return ConfigManager().get_sound_effects_tone()
+        return get_shared_config_manager().get_sound_effects_tone()
     except Exception:
         return DEFAULT_SOUND_EFFECT_TONE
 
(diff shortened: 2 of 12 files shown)


jatinkrmalik (changes requested):
@kacperpaczos Thank you for raising this. Did a quick adversarial review.

The three production constructors (`main.py`, `tray_indicator.py`, `audio_feedback.py`) now go through one lazy process-wide instance. That is the right fix for the stale-cache overwrite in #689. Settings still gets the tray's instance by injection, so it is on the same object. CI is green.

**Test nit**

`test_saves_through_the_accessor_do_not_revert_each_other` does not actually reproduce #689. You take `second` after `first.save_config()`, so a fresh `ConfigManager()` would load the already-written engine and the asserts would still pass. The identity test is what proves sharing.

Can we do this before the first mutation?

```python
first = second = get_shared_config_manager()
assert first is second
```

Even better, add a characterization test that constructs two `ConfigManager()` instances before either save and shows the second save reverts the engine. That is the footgun you left in place for tests.

**Not blocking**

`get_shared_config_manager()` is check-then-act with no lock. `ConfigManager.__init__` does `makedirs` + `load_config`, so two threads could both see `None` and leave two live instances. Not a startup problem today, because `main()` creates it on the UI thread before workers. A `threading.Lock` around the create, or documenting "UI-thread only," would close it.

`test_main.py` and `test_audio_feedback.py` still patch `vocalinux.ui.config_manager.ConfigManager`, not `get_shared_config_manager`. That only works because conftest nulls `_shared_instance` so the accessor constructs through the patched class. Prefer patching the accessor, like the tray tests now do, so a leftover singleton cannot silently skip the mock.

Otherwise this looks good to merge.

jatinkrmalik (approved):
Follow-ups look good. The accessor is locked, the tests take both handles before mutating, and CI is green.

PR #694, opened 2026-08-16
fix(tests): stop the suite from overwriting the user's real config.json

### LINKED ISSUE #693: Running the test suite overwrites the user's real config.json
## Symptom

Running the test suite (`pytest tests/`, so also `make test` / `install.sh --test`) **overwrites the developer's real `~/.config/vocalinux/config.json`**. All personal settings — engine, model, language, shortcuts, autostart — are silently replaced. After the next app start the file settles into defaults merged with a stray `general.language: "en"` key, which is the fingerprint of the synthetic fixture below.

I hit this twice on a real machine (settings reverted to whisper.cpp/tiny out of nowhere) before tracking it down.

## Diagnosis

`tests/test_config_manager.py::TestTypedAccessors` builds its manager without running `__init__`:

```python
self.config_manager = ConfigManager.__new__(ConfigManager)
self.config_manager.config = {
    "general": {"autostart": True, "language": "en"},
    ...
}
```

Unlike `TestConfigManager`, this class never patches `CONFIG_DIR`/`CONFIG_FILE` — which is fine for the accessor tests, but `test_whispercpp_advanced_persistence` then calls `save_config()`, and `save_config()` writes to the module-level `CONFIG_FILE`, i.e. the real user path. The second half of the test (`ConfigManager()`) also *reads* the real file.

Caught red-handed by making the real config read-only and running the suite with live logging:

```
tests/test_config_manager.py::TestTypedAccessors::test_whispercpp_advanced_persistence
ERROR    vocalinux.ui.config_manager:config_manager.py:239 Failed to save config:
         [Errno 13] Permission denied: '/home/paczos/.config/vocalinux/config.json'
FAILED
```

It is the only write to the real path in the entire suite.

## Impact

Everyone who runs the test suite on a machine where they also use Vocalinux loses their settings, with no error and no hint which of ~2000 tests did it. CI is unaffected (no pre-existing config worth losing), which is presumably why it went unnoticed.

## Fix

Patch `CONFIG_DIR`/`CONFIG_FILE` to a temp directory inside that one test, the same way `TestConfigManager`'s setUp already does.

kacperpaczos:
Fixes #693. Based on `main`.

## Problem

Running the test suite overwrites the developer's real `~/.config/vocalinux/config.json` with a synthetic fixture. `TestTypedAccessors` builds its manager with `ConfigManager.__new__` and never patches `CONFIG_DIR`/`CONFIG_FILE`; `test_whispercpp_advanced_persistence` then calls `save_config()`, which writes to the module-level real path (and its second half reads it back through a real `ConfigManager()`).

Caught by making the real file read-only and running the suite with live logging — this test is the only write to the real path in the entire suite; full trail in #693.

## Change

Wrap that one test in `CONFIG_DIR`/`CONFIG_FILE` patches pointing at a temp directory, the same pattern `TestConfigManager.setUp` already uses. No production code touched.

## Verification

- The test still passes and still exercises real save/load round-tripping (against the temp path).
- Running the whole `tests/test_config_manager.py` before/after: the real config's checksum previously changed; now it is byte-identical.


jatinkrmalik (approved):
@kacperpaczos Thank you for raising this. Did a quick adversarial review.

This was writing the `TestTypedAccessors` fixture over the real `CONFIG_FILE`. The patches match `TestConfigManager.setUp`, and `save_config()` plus the later `ConfigManager()` both resolve `CONFIG_FILE` at runtime, so the round-trip is still a real JSON save/load against the temp path. Production code is untouched. CI is green, including this test on Python 3.9.

One follow-up, not blocking: can we keep the path patches on the class, or move this test into `TestConfigManager`? `setUp` still uses `__new__` with no `CONFIG_DIR` / `CONFIG_FILE` redirect, so the next `save_config()` in this class will regress #693.

Otherwise this looks good to merge.

Issue #701, opened 2026-08-17
[EPIC] Build & distribution overhaul: uv, install.sh, AppImage, native packages


sesav:
## 🚀 Feature Description

Every install path Vocalinux ships (`install.sh`, PyPI, the AppImage, the AUR package, CI) resolved its dependencies at install time from unpinned ranges, with no lock file of any kind. One path installed a native ML library from an unpinned git HEAD, nothing verified the checksum of any download, and the AppImage advertised as the universal "download and run" option was built against a glibc too new for a large share of our target distros.

Most of that is fixed. `uv.lock` and hash-pinned exports landed in #705, model checksums in #713, the AppImage glibc floor in #743 and #744, the release path in #759. **v0.16.2 (2026-09-05) is the first verifiable Vocalinux release**: `SHA256SUMS` published, provenance attested on all four artifacts the run built, PyPI reached with no stored token, and the bytes on PyPI measured equal to the bytes on GitHub. It is also the first release whose PKGBUILD was built on Arch before publishing, via #772. #791 then closed the gap that release exposed, adding a checker that reads the published release rather than the workflow that produced it.

What is left splits in two.

**`install.sh` now runs in CI, and the split is what is left.** Phase 2.4 landed: the distro matrix runs `install.sh --auto --skip-models` to completion in all six containers and can fail a build. It found two openSUSE defects on its first run, in the arm that had never executed in CI. The file is still 4793 lines, up 15% since this epic opened asking for it to be split, and 2.1 is now the open half.

**`pyproject.toml` is pinned; the copies of it are not.** The Flatpak's dependency list, the PKGBUILD's `depends=()`, the Snap's build environment and `install.sh`'s install-time re-parse are four hand-maintained copies of our dependency set. No test compares any of them back to source, and three of the four have already drifted.

### Verified baseline

Re-checked 2026-09-09 against the working tree at `94c2eaa` and against the published v0.16.2 release, by querying the release, the attestations API and the secrets API. Not by reading our own workflows or docs.

[...]

PR #700, opened 2026-08-17
chore(build): harden installer script and replace Makefile with Justfile

sesav:
Ref #701 before start

This pull request replaces the `Makefile`-based development workflow with a `justfile`-based one, updates all documentation and CI references accordingly, and significantly improves the robustness and debuggability of the `install.sh` script. The installer now uses stricter error handling, structured exit codes, persistent logging, improved network checks, and more reliable dependency detection and error messages.

The most important changes are:

**Migration from Makefile to justfile:**

* All references to `make` commands in documentation (`AGENTS.md`, `docs/RELEASE_PROCESS.md`) and CI configuration (`.github/labeler.yml`, `.github/workflows/unified-pipeline.yml`) are replaced with their `just` equivalents. The `Makefile` is removed entirely.

**Installer robustness and debuggability:**

* `install.sh` now uses `set -Eeuo pipefail` for strict error handling, introduces structured exit codes for different failure scenarios, and ensures all state variables are initialized. [[1]](diffhunk://#diff-043df5bdbf6639d7a77e1d44c5226fd7371e5259a1e4df3a0dd5d64c30dca44fL5-R14) [[2]](diffhunk://#diff-043df5bdbf6639d7a77e1d44c5226fd7371e5259a1e4df3a0dd5d64c30dca44fR245-R251)
* All installer output is now logged to a timestamped file in `~/.local/state/vocalinux/`, and a scratch directory for temporary files is used and cleaned up automatically on success or preserved on failure for debugging.

**Installer network and dependency handling:**

* A new `check_connectivity` function checks HTTP(S) connectivity to required endpoints, improving reliability in restricted environments.
* The logic for resolving the release tag from the GitHub API is more robust and will not silently fall back to an outdated release; if the tag cannot be determined, the installer aborts with a clear message. [[1]](diffhunk://#diff-043df5bdbf6639d7a77e1d44c5226fd7371e5259a1e4df3a0dd5d64c30dca44fL349-R445) [[2]](diffhunk://#diff-043df5bdbf6639d7a77e1d44c5226fd7371e5259a1e4df3a0dd5d64c30dca44fR595-R599)

**Improved error reporting and user feedback:**

[...]

PR #705, opened 2026-08-18
chore(deps): lock dependencies with uv and pin every build input (epic #701, phases 0–1)

sesav:
**Base:** `feature/add-justfile` (stacked on #700 — this PR shows only the phase 0/1 commits). 

**Epic:** #701. Implements Phase 0 (dependency hygiene) and Phase 1 (uv + pinning). Does not close the epic.

## Description

Every install path (installer, PyPI, AppImage, CI) resolved dependencies at install time from unpinned ranges; there was no lock file of any kind, and the GPU installer built a native library from an unpinned git HEAD. This PR establishes pinned, hash-verified dependencies and cleans up the dependency set first so the lock doesn't freeze existing mistakes.

**Phase 0 — dependency hygiene**

- Drop `pydub`, `lxml`, `tqdm`, `python-xlib` from `dependencies` — zero imports repo-wide (`python-xlib` still arrives transitively via pynput on X11; `tqdm` legitimately returns as a transitive dep of pywhispercpp and stays in the lock).

- Add upper bounds to the risky floats: `numpy<3`, `pywhispercpp<2`, `pynput<2`, `evdev<2`, `psutil<8`.

- `vosk` moves to the optional `[vosk]` extra (both of its imports are lazy; wheel-only vosk blocks source builds while whisper.cpp is the default engine). `install.sh` installs the extra in the `--engine=vosk` branch and in the pywhispercpp-failure fallback.

- `[build-system]` tightened to `setuptools>=61,<82`, redundant `wheel` dropped.

- Fixed along the way: README advertised the removed `--with-whisper` flag (the flag parser exits 1 on it) → replaced with `--engine=` forms; the installer welcome said "Double-tap Ctrl" while writing `right_alt+right_alt` push-to-talk → now describes hold-Right-Alt; `docs/RELEASE_PROCESS.md` Step 8 now lists the AppImage and AUR jobs the tag workflow actually runs; three never-parsed installer flag variables removed.

**Phase 1 — uv lockfiles and pinned build inputs**

Two tiers, because PyGObject cannot be reliably pip-installed on user machines (needs glib 2.80+ / `girepository-2.0` + dev headers) — a plain `uv sync` would try to build it:

- **Tier A (dev/CI):** `uv.lock` (universal, py3.9–3.14, all extras — PyGObject's sdist metadata resolves without building), `.python-version` (3.13), `[tool.uv] required-version`, `just lock` / `just lock-check` targets. Installs skip PyGObject (`--no-install-package pygobject`) in favour of distro `python3-gi` via `--system-site-packages`.

- **Tier B (installer/AppImage/Flatpak):** hash-pinned exports regenerated by `just lock` — `requirements/{runtime,vad,dev}.txt` (`--no-emit-project`, `--no-emit-package pygobject`) and `requirements/whisper.txt` compiled from `whisper.in` against the PyTorch CPU index. Two findings baked into the pins:

  - PyPI's CUDA-bundled torch wheels win resolution over the CPU index regardless of index order → torch/torchaudio must be pinned to `+cpu` local versions;

  - torchaudio lags torch on the CPU index → the pair is pinned together at `2.11.0+cpu`.

- **install.sh:** `pywhispercpp` installs from `PYWHISPERCPP_VERSION="1.5.0"` (defined once) instead of unpinned `git+https://…`; the Vulkan/CUDA paths force the sdist via `--no-binary pywhispercpp`.

## Type of Change

- [x] 🐛 Bug fix (README advertised a flag the installer rejects; welcome message described the wrong hotkey)
- [x] 💥 Breaking change (for PyPI consumers: `pip install vocalinux` no longer pulls `vosk` — use `pip install "vocalinux[vosk]"`; all `install.sh` paths are handled in this PR)
- [x] 📖 Documentation update (README `--engine=` forms, RELEASE_PROCESS job list, lock-file header comments)
- [x] 🧹 Code refactoring (dead installer flags removed; pywhispercpp install source deduplicated into one pinned variable)

## Additional Notes

**Verification performed**

- Clean venv: `uv pip install -r requirements/runtime.txt -r requirements/vad.txt` (hash-checked) + the built wheel with `--no-deps` -> app imports, and` pywhispercpp.model.Model` loads with its native libs
- Two `SOURCE_DATE_EPOCH`-pinned `uv build --wheel` runs produce a byte-identical wheel (matching sha256)
- `uv lock --check` clean; `bash -n install.sh` clean


jatinkrmalik (dismissed):
Approving the uv lock / deps slice (phases 0-1 of #701). Stacked on #700, which just landed.

jatinkrmalik (approved):
Rebased onto the #700 squash. Lock/deps slice only. Landing.

PR #713, opened 2026-08-19
chore(build): uv tooling, a Python 3.11 floor, and verified model downloads (epic #701, phases 2.5 & 5)

sesav:
## Summary

Started as "move the dev workflow to uv", grew into settling the Python story the migration exposed (which interpreter the installer builds against, which versions we actually support, whether the lock means anything), and ended by closing the supply-chain gap the epic flagged. Each commit is self-contained and green on its own.

Rebased onto `main` at v0.16.0 (#718, #728, #729): the two AppImage `pywhispercpp` pin commits that started here are on main already and are no longer part of this diff.

## uv tooling and CI

Every `just` recipe runs through `uv run` against `.venv/`. Because `uv sync` prunes whatever the flags do not name, recipes share a `DEV_EXTRAS` variable; `just deps-all` was missing `--group lint` and really did uninstall the linters.

Both Python jobs now use `astral-sh/setup-uv` with `uv sync --locked` / `uv run --locked`, which also fails the build when `uv.lock` drifts from `pyproject.toml`. The first CI run failed where the migration was thinnest: `uv run --extra dev` syncs the whole project, so the lint job tried to build `pyaudio` from source. black/isort/flake8 moved into a `lint` dependency group, and CI runs `--only-group lint`, which skips the project entirely. The move had a tail: `just lock` still exported `requirements/dev.txt` from `--extra dev` alone, so the committed file reproduced neither `just deps` nor what CI lints with. It exports the group too now, and a test asserts every linter reaches the file and that no documented `uv run` example stops at `--extra dev`, which would uninstall them from `.venv`.

## Python floor -> 3.11

`requires-python` promised 3.9, the CI matrix ran 3.9–3.13, the classifiers advertised 3.14, mypy targeted 3.12 and `install.sh` enforced 3.9 — five answers to one question, and the reason an unsupported interpreter could quietly become the one users install with.

Now 3.11 everywhere, and **fatal**: the installer used to print "Continuing with unsupported Python version" and build the venv anyway. Docs, the README badge, the website and the AUR package said "3.9+" / "Ubuntu 22.04+" long after the floor moved; jammy ships python3.10 and its `python3-gi` is built for that interpreter, so it cannot satisfy the floor at all. Requirements are now Debian 12+, Ubuntu 24.04+, Fedora 39+, Arch. The distro matrix drops `ubuntu:22.04` and adds `ubuntu:26.04`.

`tests/test_python_version_policy.py` pins the invariant: installer floor == `requires-python` == lowest classifier == CI matrix, the AUR package matches, and no file under `docs/` or `web/src/` may advertise an older floor — scanned as trees, so a new page cannot reintroduce it.

## Installer correctness

- **Builds `venv/` from the system Python.** Run from a shell still holding uv's `.venv` — the state `just deps` leaves you in — it built `venv/` from that interpreter and aborted with *"Distro PyGObject is not importable in the venv"*. It now drops an inherited `VIRTUAL_ENV`, prefers `$SYSTEM_PYTHON`, and matches an existing venv by `sys.base_prefix` rather than by `X.Y`, so a venv from a different 3.12 is rebuilt instead of reused.
- **Gates on the interpreter, not on the distro release label.** `check_ubuntu_version` compared `VERSION_ID` against "24.04" for the whole Ubuntu *family*, which is `linuxmint`, `pop`, `elementary` and `zorin` as well as Ubuntu. Derivatives do not share Ubuntu's numbering: Linux Mint 22 and elementary OS 8 are built on Ubuntu 24.04 and ship Python 3.12, but report "22" and "8", which sort below "24.04", so raising that floor from 18.04 turned away systems that meet every requirement. The label check is gone; `check_python_version` was already the real gate and is fatal, so anything genuinely too old still stops, with a message about the interpreter. `install_system_dependencies` had already learned this lesson for `util-linux-extra` and probes `apt-cache` instead of the version.
- **Stops warning off distros the project documents as supported.** The compatibility block treated everything outside ubuntu/debian as "primarily designed for Ubuntu/Debian-based systems" and asked a y/n that exits on "n" — including Fedora, Arch and openSUSE, all listed as supported in `DISTRO_COMPATIBILITY.md`, and Fedora builds in the distro matrix. The tiers now mirror that document, and a test fails if the two disagree. The document needed a nudge of its own: Mint, Pop!_OS, elementary, Zorin, Manjaro and EndeavourOS were marked "Good Support" while filed under "Experimental Support", which is where the drift started. They are in the supported table now, with the base release each one needs, and a second test refuses a `✅` row in the experimental section. Nothing in the block decides whether the install proceeds; that is `check_python_version` and `require_distro_gi`.
- **Repairs a config it cannot start with.** `config.json` survives reinstalls, so it can name an engine the venv cannot import — vosk and whisper are both optional extras now. The installer rewrites it to `whisper_cpp` when that works and fails when nothing does, instead of warning and leaving a broken install.

## Model integrity

Models are 40MB–2.9GB downloads that end up in native code — whisper.cpp maps its ggml files through `ctypes.cdll`. Nothing verified any of it: `sha256|checksum|hashlib` had zero matches in `install.sh` and in the runtime downloaders.

`src/vocalinux/utils/model_checksums.txt` now pins 67 models by sha256, read by both halves (`model_checksums.py` at runtime, `awk` in `install.sh`). Verification runs **before** a file is moved into place, and **fails closed** — an unpinned model is refused, which is safe only because a test asserts the manifest covers every model the app can download.

Details worth knowing:

- **Existing files are hashed too.** All three installer functions returned early when the file was already there, so the check only ever covered a fresh download. VOSK is checked as an extracted tree, so extraction records the verified zip digest in `.vocalinux_verified`.
- **VOSK is pinned by bytes we fetched**, not the md5 Alphacephei publishes next to the download — copying that pins nothing they could not change alongside the file. The generator downloads each model once, checks the published md5 against what arrived, and stores the sha256. Incremental: only new models cost bandwidth.
- **whisper.cpp URLs use a pinned Hugging Face commit**, not `main`, or upstream replacing a file would invalidate every digest at once.
- **OpenAI Whisper is left to `openai-whisper`**, which already verifies against the sha256 in its own URL. Our downloader added no safety and two bugs: it ignored where `load_model` looks, and named `large` as `large.pt` where upstream stores `large-v3.pt`, so 2.9GB was fetched and stored twice. It now downloads only for progress and cancel; all three engines share one streaming helper.
- **An extracted tree carries a stamp, and both downloaders write it.** VOSK is the one model that arrives as an archive: the pin covers the zip, which is deleted after unpacking, so the directory left behind has no digest of its own. `install.sh` and the app both record the verified zip digest in `.vocalinux_verified` inside the tree, under a name they share (`VERIFICATION_STAMP_NAME`, asserted against `install.sh` in a test). Whoever unpacks stamps — otherwise a model fetched at first run or from Settings looks untrusted to the next `./install.sh`, which downloads it again. A tree that cannot be stamped is an error on both sides.
- **Replacing an unstamped tree is a swap, not a delete.** No stamp is not the same as known-bad — an older install, or one the app unpacked — and this runs on every install. Both halves do the same thing now: download, verify, unpack into a staging directory beside the model, stamp it there, then rename it into place, with the tree being replaced moved aside and put back if that rename fails. Staging inside the models directory keeps both renames same-filesystem, and it makes "extracted" and "stamped" one step, so a stamp that cannot be written leaves nothing an unstamped tree could be mistaken for. Deleting on sight, as the first version did, traded a working model for no model whenever the download, the network or `unzip` failed afterwards.

- **A model already on disk is hashed before it reaches native code.** Downloads are verified before their rename, but that says nothing about a file which never came through that path: releases before this one fetched ggml models from `resolve/main` unchecked, and `install.sh` re-hashes only `ggml-tiny.bin`. `_init_whispercpp` now verifies what it finds and demotes a failure to "not downloaded". This is done on every launch rather than cached, because caching it is not worth what it costs to explain: sha256 runs at ~4GB/s, which is 0.02s for the default `tiny` and 0.78s for `large`, over pages whisper.cpp is about to read in full anyway.
- **A checkpoint an earlier release misnamed is adopted, not abandoned.** That release saved OpenAI Whisper's `large` as `large.pt`, from the same large-v3 URL used today. Under the old name `load_model` cannot see it, so 2.9GB is refetched; Settings does not list it; and what is not listed cannot be deleted there either. It is renamed on the way past, which is safe rather than lucky, since the bytes already are that checkpoint and `openai-whisper` verifies whatever it loads.

## Test fixes found along the way

Local runs failed with 185 errors while CI stayed green: tests read the developer's real `~/.local/share/vocalinux` through default XDG paths and switched IBus code paths via the desktop session env. `conftest.py` redirects XDG and clears session variables at import time. Python 3.14's `unittest.mock` re-instantiates a user subclass when building child mocks, which broke the IBus engine double — replaced with a plain stand-in.

Two tests were passing vacuously: the VOSK progress test built its fixture zip at call time, when `zipfile`/`BytesIO` are already mocked, so it streamed zero bytes; and `time.time` was patched with a fixed four-value list that logging drains. Both fixed rather than muted.

One more leak, found by the first test in the suite that extracts a real archive: `TestSpeechRecognition` patched `os.makedirs` twice and stopped the two patchers in the order they were started, so unittest restored the *inner* mock into the `os` module and every test after that class ran with a fake `makedirs`. Nothing noticed while no test wrote real files. `conftest.py` also keeps a reference to the genuine `zipfile`, which two modules replace with a `MagicMock` at import time and never restore.

Also a real crash: `_get_system_model_paths()` called `paths.remove()` unconditionally and raised `ValueError` on Arch hosts with a custom `XDG_DATA_DIRS`.

## Testing

- CI green on `9fde6db`: lint, Python 3.11–3.14, AppImage smoke builds (x86_64 and aarch64), the distro matrix (Debian 12, Fedora 39, Ubuntu 24.04 and 26.04), Flatpak, website build
- 2377 tests locally, each synced from the lock; `just lint` and `uv lock --check` clean; `bash -n install.sh`
- Model verification checked against real downloads, not only mocks: the pinned digests match what the servers return, a single flipped byte is rejected, and the bash and Python implementations agree
- `install_vosk_models` exercised for real against a stub network and a fixture archive: a failed download, and a download that fails verification, both leave an existing unstamped tree untouched; a verified one replaces it and is stamped; a stamped tree is not fetched twice
- End-to-end install verified on a Debian 12 VM

## Left for #701

The **Python jobs** run from the lock — that is the part this PR delivers, not phase 5 as a whole. The AppImage, nightly and release workflows are still `setup-python` + pip, and all 12 action references are floating tags rather than SHAs. Phase 5 also still owns trusted publishing (`release.yml` declares `id-token: write` but authenticates with a token) and `sha256sums` / provenance on release artifacts. Phase 2 beyond the model checksums, and phase 3, are untouched.


install.sh
```diff
@@ -3757,6 +4018,14 @@ install_whispercpp_model() {
         return 1
     fi
 
+    # Verify the checksum before the file reaches its final location, where the
+    # "already exists" check above would treat it as a good model on every re-run.
```
jatinkrmalik: This comment is doing the work the code doesn't. If the file is already at TINY_MODEL_PATH we return 0 above and never get here. That's every re-run and every existing install. Hash it before trusting it, or delete on mismatch. Same skip is in the Whisper and Vosk downloaders.

src/vocalinux/speech_recognition/recognition_manager.py
```diff
@@ -2323,6 +2336,13 @@ def _download_whisper_model(self, cache_dir: str):
 
                         logger.info(f"Download progress: {progress * 100:.1f}% - {status}")
 
+            # OpenAI publishes each checkpoint under a path segment that is its
+            # sha256, so the expected digest arrives with the URL and needs no
+            # manifest entry. Verify before the rename, as with the other engines.
+            if self._download_progress_callback:
+                self._download_progress_callback(1.0, 0, "Verifying model...")
+            verify_openai_model(temp_file, url, description=f"{self.model_size}.pt")
```
jatinkrmalik: This only runs if we actually went through `_download_whisper_model`. `_init_whisper` still treats ~/.cache/whisper/{size}.pt as already downloaded and then calls whisper.load_model, which will fetch into download_root on its own. That path never calls this.

docs/INSTALL.md
```diff
@@ -137,7 +137,7 @@ On Ubuntu 24.04+ or Pop!_OS, install `libgirepository-2.0-dev` if
 | Requirement | Details |
 |-------------|---------|
 | **Operating System** | Ubuntu 22.04+ (recommended), Debian 11+, Fedora 38+, Arch Linux |
-| **Python** | 3.9 or newer |
+| **Python** | 3.11 or newer |
```
jatinkrmalik, line 140: The row above still says Ubuntu 22.04 recommended and Debian 11+. jammy python3 is 3.10 and python3-gi is built for that interpreter. A 3.11 venv on 22.04 cannot import the distro gi, and a 3.10 venv cannot install this package. One of these rows has to change. Website and AUR are still on 3.9+ too.

install.sh
```diff
@@ -2411,34 +2441,116 @@ mkdir -p "$DATA_DIR/models"
...
+    PYTHON_CMD="${ok_version:-$first}"
+    [ -n "$PYTHON_CMD" ]
+}
+
 # Check Python version
 check_python_version() {
-    local MIN_VERSION="3.9"
-    local PYTHON_CMD="python3"
+    # Keep in sync with requires-python in pyproject.toml.
+    local MIN_VERSION="3.11"
```
jatinkrmalik, line 2499: The function now errors, then the caller still prints "Continuing with unsupported Python version" and builds the venv anyway. On 22.04 that's the path you get. Make this fatal or the floor only exists in pyproject.

tests/test_installer_python_selection.py
```diff
@@ -0,0 +1,203 @@
...
+        assert f"SELECTED={path_python}" in self._select(path_python, system_python)
+
+    def test_falls_back_to_version_match_when_no_candidate_has_gi(self, tmp_path):
+        # gi is installed later in the run, so selection must still yield the
+        # system interpreter rather than the placeholder "python3".
+        path_python = _fake_python(tmp_path / "pathbin" / "python3", "3.13", has_gi=False)
+        system_python = _fake_python(tmp_path / "usrbin" / "python3", "3.14", has_gi=False)
+
+        stdout = self._select(path_python, system_python)
+        assert f"SELECTED={path_python}" in stdout
```
jatinkrmalik, line 155: Comment says this should yield the system interpreter. Assert wants path_python (3.13) over system 3.14. That's the gi-invisible venv bug, encoded as the expected result.

install.sh
```diff
@@ -2411,34 +2441,116 @@ mkdir -p "$DATA_DIR/models"
...
+    "$1" -c "import gi" >/dev/null 2>&1
+}
+
+# Pick the interpreter to build the venv from. Distro PyGObject is compiled for
+# exactly one Python, so prefer a candidate that can already import gi:
+# $SYSTEM_PYTHON is probed explicitly because PATH may expose a different
+# interpreter (pyenv, uv, /usr/local) that the distro packages were never built
+# for. Falls back to the newest-enough candidate when none of them has gi yet —
+# require_distro_gi() reports that case with a proper message later on.
+select_python_interpreter() {
```
jatinkrmalik, line 2460: PATH python3 is first in the candidate list, so when neither has gi yet we keep pyenv/uv instead of $SYSTEM_PYTHON. Prefer the system one when it exists and meets the floor. Also compare base_prefix, not just the X.Y string, so two different 3.12s don't get reused.

install.sh
```diff
@@ -3974,6 +4243,39 @@ else
     print_info "Models will be downloaded automatically on first application run"
 fi
 
+# config.json survives reinstalls — every writer above guards on "file does not
+# exist" — so an engine picked by an earlier attempt can outlive the venv that
+# supported it. vosk is the only engine shipped as an optional extra, so it is
+# the one that goes missing; say so here instead of letting the app die at
+# startup with ModuleNotFoundError.
+verify_configured_engine() {
```
jatinkrmalik, line 4334: Warning-only, vosk-only, and the test asserts EXIT=0. A leftover engine=whisper with no extra still dies at startup. If the point is "don't leave a broken install", rewrite the config to whisper_cpp or fail here.

scripts/generate-model-checksums.py
```diff
@@ -0,0 +1,171 @@
...
+            "Either the model was renamed upstream or whispercpp_model_info.py lists "
+            "a file that does not exist."
+        )
+    return entries, revision
+
+
+def _vosk_entries(wanted: Iterable[str]) -> list[Entry]:
+    payload = _fetch_json(VOSK_MODEL_LIST)
+    by_name = {
+        item["name"]: (item["md5"], int(item.get("size", 0))) for item in payload if item.get("md5")
```
jatinkrmalik: We're pinning whatever Alphacephei put in model-list.json, not a hash of a file we fetched. MD5 is also the one we'd pick last. If this is the supply-chain pin, download the zip once in this script and store sha256 of the bytes.

justfile
```diff
@@ -70,6 +94,11 @@ lock:
 lock-check:
     uv lock --check
 
+# Refresh the pinned model digests from upstream metadata (no models downloaded).
+# Run after adding a model to vosk_model_info.py or whispercpp_model_info.py.
+model-checksums:
+    uv run python scripts/generate-model-checksums.py
```
jatinkrmalik: Header of this file is about uv run dropping extras when a recipe forgets DEV_EXTRAS. This recipe forgot. deps-all has the same hole (uv sync --all-extras, no lint group).

.github/workflows/unified-pipeline.yml
```diff
@@ -70,28 +70,26 @@ jobs:
     steps:
       - uses: actions/checkout@v5
 
-      - name: Set up Python
-        uses: actions/setup-python@v5
+      - name: Set up uv
+        uses: astral-sh/setup-uv@v6
```
jatinkrmalik, line 74: Lint/test jobs are --locked, which is the useful part. AppImage, nightly, and release are still setup-python + pip, and this action is a floating tag. I'd say "python jobs use the lock" in the PR text rather than calling it phase 5.

AGENTS.md
```diff
@@ -5,7 +5,7 @@ Guidelines for AI agents working on this codebase.
 ## Project Overview
```
jatinkrmalik: This file is the merge conflict against current main. Main has the worktree rules at the top. This branch still has the old Project Overview block. Rebase and keep main's structure.

tests/test_python_version_policy.py
```diff
@@ -0,0 +1,79 @@
...
+
+import re
+import sys
+import tomllib
+from pathlib import Path
+
+REPO_ROOT = Path(__file__).resolve().parents[1]
+PYPROJECT = REPO_ROOT / "pyproject.toml"
+INSTALL_SH = REPO_ROOT / "install.sh"
+PIPELINE = REPO_ROOT / ".github" / "workflows" / "unified-pipeline.yml"
```
jatinkrmalik, line 16: This only looks at pyproject, install.sh, classifiers, and the unified-pipeline matrix. Website (page.tsx, FAQ, distro install pages) and packaging/aur still say 3.9+. If the test is meant to stop the floor drifting, those need to be in here too.

jatinkrmalik (changes requested):
@sesav I left a bunch of comments on the diff, please take a look.

I like the idea, it's a solid idea.

AGENTS.md
sesav: You're right, I made a mistake during the rebase. The file has been completely revised.

install.sh
sesav: You're right, and the comment was the worst part of it, promising something the code didn't do. Fixed in d43bd21.

`tiny.pt` and `ggml-tiny.bin` are hashed on the exists-path now, a mismatch deletes the file and re-downloads. VOSK needed a different shape, since the check is on the extracted tree and the zip the digest covers is gone by then — extraction writes the verified digest to `.vocalinux_verified`, and the tree is trusted only while that stamp matches what we pin. Older installs and manual copies have no stamp, so they get re-downloaded.

Tests assert each guard verifies before its `return 0`; I checked they fail against the old behaviour rather than passing either way. 🫡

One I left alone: the runtime skips the same way (`is_model_downloaded()` is just `os.path.exists()`). Hashing up to 2.9GB per launch isn't viable, so it probably wants the same stamp trick — here, or a follow-up?

src/vocalinux/speech_recognition/recognition_manager.py
sesav: Good catch! And it went further than the verification. `load_model()` already verifies what it loads, from the same sha256 in the URL, so our check there was redundant. The downloader also had two bugs of its own: it ignored where `load_model` actually looks, and named `large` as `large.pt` instead of `large-v3.pt`, so that model was fetched and stored twice.

Changed: Vocalinux now downloads only for progress and cancel and leaves verification to `load_model`. All three engines share one streaming helper, so VOSK's duplicate loop is gone. `install.sh` has one download function instead of three; and the file name comes from the URL basename in one place.

docs/INSTALL.md
sesav: > jammy python3 is 3.10 and python3-gi is built for that interpreter

Correct, and I'd only tested on Debian 12, which ships 3.11.

> A 3.11 venv on 22.04 cannot import the distro gi, and a 3.10 venv cannot
> install this package

Right. Supporting jammy would mean holding the whole dependency set to an interpreter that's nearly EOL, with its own test pipeline and a .deb to distribute. Two LTS releases have shipped since. Not worth it.

> One of these rows has to change.

Done! And it was in nine more places than `INSTALL.md`: DISTRO_COMPATIBILITY, the README badge, CONTRIBUTING, the AUR PKGBUILD, and five website files still saying "Python 3.9+" / "Ubuntu 22.04+". install.sh's own `check_ubuntu_version` floor was still 18.04.

Requirements are now Debian 12+, Ubuntu 24.04+, Fedora 38+, Arch. The distro matrix drops ubuntu:22.04 and adds ubuntu:26.04.

To stop it drifting again, test_python_version_policy now also asserts the AUR package matches requires-python and that no doc or website file advertises a Python below the floor which covers your other comment on that test. Both checks fail against the state before this change.

install.sh
sesav: That was still warn-and-continue, I'd only touched the adjacent `check_ubuntu_version` floor, not this one.

It exits now. `check_python_version` already prints the detected version, so the call site just adds the way out and aborts with EXIT_MISSING_DEPS:

      Point SYSTEM_PYTHON at a newer interpreter if one is installed:
        SYSTEM_PYTHON=/usr/bin/python3.12 ./install.sh

Added test_python_version_policy for it, and fails against the previous behaviour.

With 22.04 dropped from the docs and the distro matrix in the same push, the floor is now the same number in pyproject, `install.sh` and the CI matrix.

tests/test_installer_python_selection.py
sesav: Right, and the test was the smaller half of it, the comment above the assert was correct and the assert was wrong. 🙁

The candidate list is PATH python3 first, then $SYSTEM_PYTHON. That only matters when nothing has gi yet, which is exactly first-run, since install.sh installs python3-gi later in the same run. 

The fallback now prefers $SYSTEM_PYTHON when it meets the floor. gi still wins when a candidate has it, so a gi-capable PATH interpreter is still selected.

Fixed the assert, and added a case for the system interpreter being below the floor, where the PATH one has to win.

install.sh
sesav: Both parts fixed.

Candidate order: PATH python3 came first, which only mattered when nothing had gi yet — i.e. first run, since install.sh installs python3-gi later in the same run. The fallback took the first version-eligible candidate, so it kept the uv .venv 3.13 over the system 3.14. It now prefers $SYSTEM_PYTHON when that meets the floor; gi still wins outright when a candidate has it, and there's a test for the system interpreter being below the floor, where the PATH one has to win.

base_prefix: `venv_matches_selected_python` compared only X.Y, so a venv built from a different 3.12 (pyenv, uv, /usr/local) counted as a match and got reused. It compares `sys.base_prefix` now — for a venv that's the interpreter it was built from, which is what decides whether distro gi is importable. Verified on real venvs here: one from /usr/bin/python3 still matches and is reused, a uv 3.13 one is recreated. The stub interpreter in the tests answers the base_prefix probe now, so two stubs are distinct installations unless told to share one.

install.sh
sesav: All three, yes.

Engines now come from a table, so `whisper` is checked too, it's an optional extra just like vosk, and a leftover `engine=whisper` died at startup the same way. `remote_api` needs nothing and is skipped.

And it repairs instead of warning: if the configured engine can't be imported but whisper_cpp can, config.json is rewritten to whisper_cpp (other settings preserved) and the install continues. If nothing works, the install fails and the config is left as-is so you can see what it asked for.

Tests cover repair-from-vosk, repair-from-whisper, and the no-working-engine failure. All three fail against the previous behaviour.

scripts/generate-model-checksums.py
sesav: Fair enough, copying their md5 pins nothing they couldn't change alongside the file. 

The generator downloads each VOSK model once now and pins the sha256 of the bytes it received. The published md5 is checked against them on the way through, so a listing that disagrees with its own file stops generation instead of pinning either. The manifest is 67 sha256 entries, no md5 anywhere, and the now unreachable md5 branch in install.sh is gone.

Downloading is incremental — a model already pinned by sha256 isn't fetched again, so only new models cost bandwidth (--refresh forces all). The first run was 21.9 GB / ~30 min, which seemed worth it once. 🫡

Left whisper.cpp as it was: HF's lfs.oid is the content address of the blob rather than a metadata field, and it's pinned to a commit.

justfile
sesav: Both fixed! `model-checksums` takes DEV_EXTRAS now, `deps-all` adds `--group lint`, and `deps` uses the variable instead of repeating it.

Worth noting the two aren't the same failure. `deps-all` really was broken: `uv sync --extra dev --extra vad` (no lint group) uninstalls 8 packages here, so that recipe left you without linters. But `uv run` on 0.12.5 doesn't prune, I ran the old recipe and all four tools survived, no "uninstalled" line in the verbose log. So the header's claim that a recipe forgetting DEV_EXTRAS would uninstall the dev tools isn't true for current uv. 

I've reworded it to say sync prunes, run has in past versions, and recipes ask for the set so behaviour doesn't depend on which uv is installed. 👌

tests/test_python_version_policy.py
sesav: Already extended when I fixed `INSTALL.md`, and now a bit further.

The AUR package is checked against requires-python, and no doc or website file may advertise a Python below the floor. That covered the files you named, but by a hardcoded list which is the same shape of gap, since a new page could reintroduce "3.9+" without the test noticing. It scans the trees now (`docs/**/*.md`, `web/src/**/*.{ts,tsx}`, README, CONTRIBUTING): 62 files today, no false positives, and a freshly added page claiming 3.9+ does fail it.

Not covered: distro claims like "Ubuntu 22.04+" — the scan only looks at Python versions. Those were fixed by hand in the same push.

.github/workflows/unified-pipeline.yml
sesav: Agreed on all three, and the PR text now says "the Python jobs run from the lock" instead of claiming phase 5.

Your second point stopped being theoretical while this was in review: `packaging/appimage/build.sh` installed `pywhispercpp` from source with no version, 1.5.1 landed on PyPI that day and failed to compile with Vulkan, and both AppImage jobs went red with nothing in the repo changed. The version was pinned in pyproject, `uv.lock` and `install.sh` everywhere except the script that builds the artifact we ship. `build.sh` reads the pin from `install.sh` now rather than repeating it, and a test fails if that reference goes away.

Still open and listed as such: nightly and release are `setup-python` + pip (2 and 3 uses), and all 12 action refs are floating tags.

install.sh
```diff
@@ -3614,10 +3893,23 @@ install_vosk_models() {
...
-        return 0
+        local EXPECTED_ZIP_DIGEST
+        EXPECTED_ZIP_DIGEST=$(pinned_digest_for "$SMALL_MODEL_ARCHIVE")
+        if [ -n "$EXPECTED_ZIP_DIGEST" ] && [ -f "$SMALL_MODEL_STAMP" ] &&
+           [ "$(cat "$SMALL_MODEL_STAMP" 2>/dev/null)" = "$EXPECTED_ZIP_DIGEST" ]; then
+            print_info "Small VOSK model already exists at $SMALL_MODEL_PATH (verified)"
+            return 0
+        fi
+        print_warning "The existing VOSK model was not verified against the pinned digest; replacing it."
+        rm -rf "$SMALL_MODEL_PATH"
```
jatinkrmalik: Stamp missing or wrong and we `rm -rf` the tree, then we check wget/curl, connectivity, and unzip. Those can fail. This runs on every install unless `--skip-models`, and no stamp is the common case: old installs, plus anything the app extracted (see the runtime comment).

whisper.cpp hashes in place and only deletes on mismatch. Same shape here: fetch and verify a replacement, then swap. If that fails, leave the tree. Missing stamp means try again when we can, not delete now.

src/vocalinux/speech_recognition/recognition_manager.py
```diff
@@ -2101,244 +2139,58 @@ def cancel_download(self):
...
                 self._download_progress_callback(1.0, 0, "Extracting model...")
-
-            # Extract the model
             logger.info(f"Extracting VOSK model to {model_path}")
             with zipfile.ZipFile(zip_path, "r") as zip_ref:
                 zip_ref.extractall(MODELS_DIR)
 
-            # Remove the zip file
             os.remove(zip_path)
             logger.info("VOSK model downloaded and extracted successfully")
```
jatinkrmalik, line 2219: Verified zip, extract, `os.remove(zip_path)`, no `.vocalinux_verified`. That's the only signal `install_vosk_models` trusts. This isn't the launch-time `exists()` skip. The zip was already hashed here and writing 64 bytes is free.

A first-run / Settings download is always unverified to the installer, so the next `./install.sh` hits the delete-first path. After extract, write `expected_for(f"{model_name}.zip").digest` to the tree, same contents as `pinned_digest_for`. Fail if the stamp can't be written. Assert it in the VOSK download test.

tests/test_recognition_manager_downloads.py
```diff
@@ -285,10 +322,68 @@ def test_stream_model_download_eta_minutes_and_empty_chunks(self, tmp_path):
...
+                    manager._download_whisper_model(str(tmp_path))
+
+        assert not list(tmp_path.glob("*"))
+
+
 class TestDownloadVoskModel:
     """Test _download_vosk_model() with runtime import mocking."""
 
-    def test_download_vosk_progress_callback(self, tmp_path):
+    def test_download_vosk_progress_callback(self, tmp_path, skip_checksum):
```
jatinkrmalik, line 437: `skip_checksum` is here so the synthetic zip isn't fail-closed, then we never assert it was called. The whisper.cpp sibling does `assert_called_once()`. Drop `verify_model_file` from `_download_vosk_model` and this still passes.

justfile
```diff
@@ -70,6 +95,11 @@ lock:
 lock-check:
     uv lock --check
 
+# Refresh the pinned model digests from upstream metadata (no models downloaded).
```
jatinkrmalik: Comment still says no models downloaded. After the VOSK change, an unpinned or `--refresh` run fetches the zips (~21.9 GB / ~30 min on first generate). `generate-model-checksums.py` says the same at the top, then the VOSK bullet contradicts it. Say VOSK zips download unless already pinned; whisper.cpp still uses HF `lfs` metadata.

src/vocalinux/utils/model_checksums.py
```diff
@@ -0,0 +1,198 @@
...
+Vocalinux pulls 40MB-2GB model files from Hugging Face, Alphacephei and
+openaipublic.azureedge.net, then feeds them to native code -- whisper.cpp maps
+its ggml files through ``ctypes.cdll``. A tampered or truncated model is
+therefore not merely a bad transcription, so every download is checked against a
+digest pinned in this repository before it is moved into place.
+
+Two sources of truth, no network lookup in either case:
+
+* ``model_checksums.txt`` next to this module pins a digest per file name for
+  whisper.cpp (sha256, from the pinned Hugging Face revision) and VOSK (md5, the
```
jatinkrmalik: Docstring still says we pin "VOSK (md5, the only hash Alphacephei publishes)". The committed manifest and `test_manifest_entries_are_well_formed` are sha256-only. Describe them as sha256 of the bytes the generator downloaded.

README.md
```diff
@@ -26,7 +26,7 @@
 [![Follow on X](https://img.shields.io/badge/Follow%20%40vocahq-000000?style=flat&logo=x&logoColor=white)](https://x.com/vocahq)
 
 <!-- Distros (widest; base plate above the hero) -->
-[![Ubuntu](https://img.shields.io/badge/Ubuntu-22.04+-E95420?logo=ubuntu&logoColor=white)](docs/DISTRO_COMPATIBILITY.md)
+[![Ubuntu](https://img.shields.io/badge/Ubuntu-24.04+-E95420?logo=ubuntu&logoColor=white)](docs/DISTRO_COMPATIBILITY.md)
 [![Debian](https://img.shields.io/badge/Debian-11+-A81D33?logo=debian&logoColor=white)](docs/DISTRO_COMPATIBILITY.md)
```
jatinkrmalik: Badge is still `Debian-11+`. Requirements, `DISTRO_COMPATIBILITY`, `INSTALL.md`, and the site are Debian 12+. Debian 11 ships 3.9, below the floor. The version-policy test only looks at `Python X.Y+` so this won't fail CI. Change the badge to `Debian-12+`.

install.sh
```diff
@@ -3974,6 +4263,97 @@ else
     print_info "Models will be downloaded automatically on first application run"
 fi
 
+# config.json survives reinstalls — every writer above guards on "file does not
+# exist" — so an engine picked by an earlier attempt can outlive the venv that
+# supported it. vosk is the only engine shipped as an optional extra, so it is
```
jatinkrmalik: This still says vosk is the only optional extra that goes missing. The table and tests handle whisper the same way and skip `remote_api`. Next paragraph already has the fallback. Drop the vosk-only sentence.

jatinkrmalik (changes requested):
@sesav Python floor, interpreter selection, engine repair, VOSK sha256, and extras look good on this head.

The stamp scheme still deletes first, and the runtime never writes `.vocalinux_verified`, so the next `./install.sh` treats a good model as untrusted. A few comments/docs still describe the old version.

install.sh
sesav: Done, in `76e3cd5`. The guard no longer deletes anything. A missing stamp only warns and falls through to the download. The tail of the function is now the shape you described. :ok_hand: 

Also added `TestVoskInstallerReplacesRatherThanDeletes` which runs the real function with `download_model_file`, `check_connectivity` and `curl` stubbed and a fixture archive standing in for the model:


- a failed download leaves an unstamped tree byte-for-byte intact;
- so does a download that fails verification;
- a verified download replaces the tree (and does not merge into it) and stamps it;
- a stamped tree is not downloaded a second time;
- no scratch directories are left behind.

I checked they actually bite: putting the `rm -rf` back makes the first two fail.

src/vocalinux/speech_recognition/recognition_manager.py
sesav: Fixed in `64b79c4`, and the shape changed again in `f865b3c` after you pointed out that writing the stamp after `extractall` is not atomic. Updating this so the thread matches the code.

`_download_vosk_model` now calls `_unpack_vosk_model`, which unpacks into `.<model>.incoming` inside `MODELS_DIR`, stamps the extracted tree there, and only then renames it into place; the tree being replaced moves to `.<model>.replaced` and is put back if that rename fails. So "extracted" and "stamped" are one step, and a stamp that cannot be written leaves nothing behind for `_init_vosk` to mistake for an installed model.

The writer itself lives in `model_checksums.py`, next to the manifest it reads, and the file name is the shared constant `VERIFICATION_STAMP_NAME`, with a test asserting `install.sh` uses the same literal.

Tests in `TestDownloadVoskModel`: the stamp exists and equals the pinned digest, a failed stamp leaves neither a tree nor scratch behind, a failed unpack keeps the model that was already installed, an archive whose top-level directory is not the model name is refused, a verified download replaces rather than merges, and verification itself is asserted.

tests/test_recognition_manager_downloads.py
sesav: Agree. Added `skip_checksum.assert_called_once()`, matching the whisper.cpp sibling.

Two things fell out of making the test mean something:

- the fixture used `model-en-us-0.22-lgraph`, which is not a VOSK model at all (the real one is `vosk-model-small-en-us-0.15`). Nothing in the manifest could match it, so the stamp had nothing to look up. The map in `_make_manager` now holds the names `_init_vosk()` really picks for en-us, and the embedded zip unpacks to that directory.
- with extraction actually asserted, the test failed in the full suite but passed alone. Two causes, both pre-existing:

  - `TestSpeechRecognition.setUp` patches `os.makedirs` twice (`mockMakeDirs` and `patcher_makedirs`) and `tearDown` stops them in start order, so unittest restored the *inner* mock into the `os` module — every test after that class ran with a fake `makedirs`. Removed the duplicate patcher.
  
  - `test_recognition_manager.py` and `test_recognition_manager_download_buffer.py` assign a `MagicMock` to `sys.modules["zipfile"]` at import time and never restore it, so `extractall` was a silent no-op. `conftest.py` now keeps the real module reachable, the way it already does for numpy.

Neither was visible before because no test in the suite extracted a real archive. 🥹

justfile
sesav: Fixed in `9fe751f`, in both places.

The justfile comment now reads:

```
# Refresh the pinned model digests. whisper.cpp digests come from Hugging Face
# `lfs` metadata and cost no bandwidth; VOSK is pinned by the bytes we fetch, so
# every zip not already in the manifest is downloaded. A first run, or any run
# with --refresh, pulls ~21.9GB and takes ~30 minutes.
```

The script docstring lost its "built without downloading a single model" lead-in. It now says where each hash comes from *and* what generating it costs, so the VOSK bullet no longer contradicts the paragraph above it: whisper.cpp resolves from `lfs.oid` with nothing downloaded, VOSK is fetched and hashed, incrementally, and a first run or `--refresh` pulls every zip.

21.9GB is the sum of the `size` column over the VOSK entries in the committed manifest, so the number moves with the manifest rather than being a guess.

README.md
sesav: Fixed in `9fe751f`. Everything else in the README already said Debian 12+; the badge was the last one.

You are right that the version-policy test only looks for `Python X.Y+`, so it would not have caught this. That gap had one more victim: `docs/INSTALL.md` still claimed **Fedora 38+** while the badge, `README:268`, `DISTRO_COMPATIBILITY.md`, three pages of the site and the distro matrix all say 39+. Fixed in `c6acf21`, so every distro floor in the tree now agrees with what CI actually tests.

Teaching the test about distro floors means encoding "distro release -> Python version" somewhere, which I would rather do alongside the `DISTRO_COMPATIBILITY` work than bolt on here. Say the word if you want it in this PR instead.

src/vocalinux/utils/model_checksums.py
sesav: Fixed in `9fe751f`.

It now reads "a sha256 per file name, for whisper.cpp (the ``lfs.oid`` at the pinned Hugging Face revision) and for VOSK (the hash of the bytes the generator downloaded; the md5 Alphacephei publishes beside a model pins nothing they could not change alongside it)".

While there: the paragraph promised "Two sources of truth" and a bullet list, but the second bullet had been rewritten into a plain paragraph that still said "absent from both". One source of truth now, and the OpenAI Whisper note reads as its own paragraph.

install.sh
sesav: Dropped in `9fe751f`. 

The comment now stops at "an engine picked by an earlier attempt can outlive the venv that supported it" and points at the repair, which is what the table and the next paragraph already describe for every engine.

jatinkrmalik (approved):
Looks good to me. Thanks @sesav for turning this around so quickly. Merging now; leftover items will go in a follow-up.

PR #714, opened 2026-08-20
fix(actions): send real BackSpace key events for "delete that"

eiseleb47:
## Description

`_handle_delete_last` built its deletion as `"\b" * len(text)` and passed it to `inject_text`, treating U+0008 as typed input. No backend delivers that as a keypress:

- `ydotool type` has no keymap entry for U+0008, so it types nothing.
- IBus `commit_text()` commits it as a literal control character, which the client inserts or drops rather than treating as a deletion.

Only an actual key event deletes, so the "delete that" voice command did nothing on those paths while still reporting success.

This adds `press_backspace(count)`, which sends `count` real BackSpace events using the tool that fits the environment: `xdotool key --repeat` on X11 and the xdotool-backed Wayland path, `wtype -k BackSpace` or the ydotool keycode pair `14:1 14:0` otherwise. WAYLAND_IBUS resolves a virtual-keyboard tool of its own, since IBus cannot send key events.

`inject_text` is untouched; only the delete_last action changes.

## Type of Change

- [ ] 📖 Documentation update
- [ ] 🧹 Code refactoring
- [ ] ✅ Test update

## Checklist

- [ ] I have updated the documentation accordingly — no user-facing docs describe this behaviour
- [x] Pre-commit hooks pass

## Additional Notes

The two existing delete_last tests asserted the U+0008 string was injected, so they now pin the new contract instead: `press_backspace` called once with the character count, and `inject_text` not called at all.

---

### DIFF
--- a/src/vocalinux/text_injection/text_injector.py
+++ b/src/vocalinux/text_injection/text_injector.py
@@ -1977,6 +1977,115 @@ class TextInjector:
             raise ValueError("no key found")
         return steps
 
+    def press_backspace(self, count: int) -> bool:
+        """Send ``count`` real BackSpace key events.
+
+        Deleting text cannot be done by injecting U+0008 characters: ydotool has
+        no keymap entry for it (so it types nothing) and IBus ``commit_text()``
+        commits it as a literal control character. Only actual key events delete.
+
+        Args:
+            count: Number of backspaces to send.
+
+        Returns:
+            True if the presses were delivered, False otherwise.
+        """
+        if count <= 0:
+            return True
+
+        logger.debug(f"Sending {count} backspace key event(s)")
+
+        if (
+            self.environment == DesktopEnvironment.X11
+            or self.environment == DesktopEnvironment.WAYLAND_XDOTOOL
+            or self.environment == DesktopEnvironment.X11_IBUS
+        ):
+            env = os.environ.copy()
+            if self.environment == DesktopEnvironment.WAYLAND_XDOTOOL:
+                env["GDK_BACKEND"] = "x11"
+                env["QT_QPA_PLATFORM"] = "xcb"
+                if not env.get("DISPLAY"):
+                    env["DISPLAY"] = ":0"
+            try:
+                # --clearmodifiers already neutralises a held modifier here, so
+                # this path needs no _wait_for_modifiers_released().
+                subprocess.run(
+                    ["xdotool", "key", "--clearmodifiers", "--repeat", str(count), "BackSpace"],
+                    env=host_env(env),
+                    check=True,
+                    stderr=subprocess.PIPE,
+                    text=True,
+                    timeout=max(3, count * 0.05),
+                )
+                return True
+            except (
+                subprocess.CalledProcessError,
+                subprocess.TimeoutExpired,
+                FileNotFoundError,
+            ) as e:
+                logger.error(f"xdotool backspace error: {e}")
+                return False
+
+        # Wayland (including WAYLAND_IBUS -- IBus cannot send key events).
+        # A held PTT modifier would turn every BackSpace into Ctrl+BackSpace
+        # (delete-word), so wait it out before resolving and sending.
+        self._wait_for_modifiers_released()
+        tool = self._resolve_wayland_key_tool()
+        if not tool:
+            logger.error("No virtual-keyboard tool available to send backspaces")
+            return False
+
+        if tool == "ydotool" and not self._ensure_ydotoold():
+            # Checked on every call, not just when the tool is first resolved:
+            # wayland_tool is cached from startup, so a daemon that died since
+            # would otherwise never be restarted for this path alone. Both other
+            # ydotool call sites warn and continue the same way.
+            logger.warning("ydotoold not ready before backspace injection")
+
+        if tool == "wtype":
+            cmd = ["wtype"] + ["-k", "BackSpace"] * count
+            timeout = max(3, count * 0.01)
+        elif self._ydotool_uses_legacy_named_keys():
+            # 0.1.x: N named tokens in one call. Its per-call delay budget is
+            # spread across all events, so this stays flat regardless of count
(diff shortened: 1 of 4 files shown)


src/vocalinux/text_injection/text_injector.py
```diff
@@ -1678,6 +1678,74 @@ def _inject_shortcut_with_wayland_tool(self, shortcut: str) -> bool:
...
+        if not tool:
+            logger.error("No virtual-keyboard tool available to send backspaces")
+            return False
+
+        if tool == "wtype":
+            cmd = ["wtype"]
+            for _ in range(count):
+                cmd += ["-k", "BackSpace"]
+        else:
+            cmd = ["ydotool", "key"] + ["14:1", "14:0"] * count
```
jatinkrmalik: @eiseleb47 Same dialect footgun as #715. `14:1 14:0` is the 1.x form. On apt 0.1.x that types garbage, still returns True, and clears `last_injected_text`, so the command cannot recover.

Please branch on `_ydotool_uses_legacy_named_keys()` the way #734 does for clipboard paste.

src/vocalinux/text_injection/text_injector.py
```diff
@@ -1678,6 +1678,74 @@ def _inject_shortcut_with_wayland_tool(self, shortcut: str) -> bool:
...
+                logger.error(f"xdotool backspace error: {e}")
+                return False
+
+        # Wayland (including WAYLAND_IBUS -- IBus cannot send key events).
+        tool = getattr(self, "wayland_tool", None)
+        if not tool:
+            tool = (
+                "wtype"
+                if shutil.which("wtype")
+                else ("ydotool" if shutil.which("ydotool") else None)
```
jatinkrmalik: @eiseleb47 When `wayland_tool` is unset this prefers the compositor helper over the uinput one. `_check_dependencies` does the opposite.

Please match that order, call `_wait_for_modifiers_released()` before sending keys, and call `_ensure_ydotoold()` when you pick the uinput helper. Held PTT modifiers turn BackSpace into delete-word, and Flatpak 1.x needs the daemon.

src/vocalinux/text_injection/text_injector.py
```diff
@@ -1678,6 +1678,74 @@ def _inject_shortcut_with_wayland_tool(self, shortcut: str) -> bool:
...
+
+        if tool == "wtype":
+            cmd = ["wtype"]
+            for _ in range(count):
+                cmd += ["-k", "BackSpace"]
+        else:
+            cmd = ["ydotool", "key"] + ["14:1", "14:0"] * count
+
+        try:
+            subprocess.run(cmd, check=True, stderr=subprocess.PIPE, text=True)
```
jatinkrmalik: @eiseleb47 Please put a timeout on this `subprocess.run`. A hung tool freezes the action thread; the inject path already treats this as untrusted.

jatinkrmalik (changes requested):
@eiseleb47 Thanks for raising this. Real BackSpace events are the right fix for "delete that", and leaving `inject_text` alone is the right scope.

I still need a few changes before this can land. Main now has a v0/v1 dialect probe from #734; this branch still emits 1.x-only codes. Please also match the existing inject path for tool order, held-modifier wait, and the daemon helper. Land order once #715 is fixed: #715, then #716, then this.

src/vocalinux/text_injection/text_injector.py
eiseleb47: Fixed — branches on `_ydotool_uses_legacy_named_keys()` now. 1.x reuses `_KEYCODES["backspace"]` rather than a second hardcoded `14`; 0.1.x emits N named `backspace` tokens in a single call.

On the recovery half of your point: `last_injected_text` is already safe, since `_handle_delete_last` clears it only inside `if success:` and `test_delete_last_failure` asserts it survives a `False` return. So the dialect fix closes this at source — the command stops reporting success while typing garbage — and I've left `action_handler.py` alone. Happy to add more there if you'd rather.

src/vocalinux/text_injection/text_injector.py
eiseleb47: All three done.

**Order** — now calls `_resolve_wayland_key_tool()`, added in #716, which mirrors `_check_dependencies`. Neither PR carries its own copy of the order any more.

**Wait** — added on the Wayland path, before the tool is resolved. The X11 branch deliberately skips it because `--clearmodifiers` already handles a held modifier there; `test_x11_uses_xdotool_repeat_without_waiting` asserts that, so it's a pinned decision rather than an omission.

**Daemon** — this one bit me, so flagging the detail. Relying on `_resolve_wayland_key_tool()` for the check wasn't enough: it early-returns when `wayland_tool` is already cached, and `_check_dependencies` caches it at startup, so `_ensure_ydotoold()` would have run once per process. Both other ydotool call sites re-check every invocation, meaning paste and shortcuts recover from a dead `ydotoold` but "delete that" wouldn't have — the exact Flatpak scenario you described. It's now an unconditional call in `press_backspace`, with a test that fails if it's removed.

src/vocalinux/text_injection/text_injector.py
eiseleb47: Added, on both branches, along with `env=host_env()`.

One adjustment to the literal ask: the timeout **scales with `count`** rather than being flat. ydotool 1.x sleeps `--key-delay` after every token (`Client/tool_key.c`, 12ms default) and N backspaces is 2N tokens, so deleting a 400-character transcription takes ~9.6s — a small fixed timeout would abort exactly the deletions that need the budget. The 1.x path now also passes `--key-delay 2`, matching what `ydotool type` already does via `VOCALINUX_YDOTOOL_KEY_DELAY`, which brings the worst case to ~1.6s against a ~4.8s budget.

The 0.1.x path is a flat 5s: it spreads one delay budget across all events, so it's ~200ms regardless of count.

PR #715, opened 2026-08-20
fix(wayland): drive keyboard shortcuts with wtype and ydotool

eiseleb47:
## Description

`_inject_shortcut_with_wayland_tool` could not deliver a shortcut on either backend:

- **wtype** was refused outright with "Keyboard shortcuts not supported with wtype", but wtype does chord: `-M` presses a modifier, `-k` types a key, `-m` releases it. Modifiers must be released explicitly, as wtype only drops them when the process exits.
- **ydotool** was handed the shortcut string as-is, `ydotool key ctrl+z`. `ydotool key` takes raw evdev keycodes (`29:1 44:1 44:0 29:0`) and documents non-interpretable values as only causing a delay, so this was a silent no-op that still returned `True`.

This parses the shortcut into a list of `(modifiers, key)` steps and emits the right form per tool. Tokens are classified by name rather than by position, because the mappings in `ActionHandler._SHORTCUT_ACTIONS` are not all modifiers-first: `select_line` is `"Home+shift+End"`, meaning press Home, then Shift+End — two sequential steps, not one chord.

An unmappable ydotool key now fails loudly instead of emitting a partial sequence, and a trailing modifier with no key is rejected at parse time.

The X11 and WAYLAND_XDOTOOL paths are untouched — xdotool already accepts the string form directly.

## Verified

On Hyprland 0.56.2 with wtype 0.4 and ydotool 1.0.4, injecting into a GTK3 window that logs key-press events:

| command | delivered |
|---|---|
| `wtype -M ctrl -k z -m ctrl` | `z` with control modifier |
| `wtype -k Home -M shift -k End -m shift` | `Home` (no mods), then `End` with shift |
| `ydotool key 29:1 44:1 44:0 29:0` | `Control_L`, then `z` with control |
| `ydotool key ctrl+z` (old form) | **nothing — exit 0, zero key events** |

## Type of Change

- [ ] 📖 Documentation update
- [ ] 🧹 Code refactoring
- [ ] ✅ Test update

## Checklist

- [x] I have updated the documentation accordingly — no user-facing docs describe this behaviour
- [x] Pre-commit hooks pass

## Additional Notes

`test_wtype_not_supported` asserted the old refusal, so it is replaced by tests pinning the emitted argv for both tools, the two-step case, and the unmappable-key rejection.

---

### DIFF
--- a/src/vocalinux/text_injection/text_injector.py
+++ b/src/vocalinux/text_injection/text_injector.py
@@ -1728,22 +1728,209 @@ class TextInjector:
         Returns:
             True if successful, False otherwise
         """
-        if self.wayland_tool == "wtype":
-            # wtype doesn't support key combinations directly, so we can't implement this easily
-            logger.warning("Keyboard shortcuts not supported with wtype")
+        try:
+            steps = self._parse_shortcut(shortcut)
+        except ValueError as e:
+            logger.error(f"Cannot parse shortcut '{shortcut}': {e}")
             return False
+
+        # A held toggle/PTT modifier rewrites the chord we are about to send, so
+        # wait it out first -- same reason and placement as _inject_with_wayland_tool.
+        self._wait_for_modifiers_released()
+
+        if self.wayland_tool == "wtype":
+            # wtype does support combinations: -M presses a modifier, -k types a
+            # key, -m releases it. Modifiers must be released explicitly -- wtype
+            # only auto-releases them when the process exits.
+            cmd = ["wtype"]
+            for modifiers, key in steps:
+                for mod in modifiers:
+                    cmd += ["-M", self._WTYPE_MODIFIERS[mod]]
+                cmd += ["-k", key]
+                for mod in reversed(modifiers):
+                    cmd += ["-m", self._WTYPE_MODIFIERS[mod]]
         elif self.wayland_tool == "ydotool":
-            try:
-                cmd = ["ydotool", "key", shortcut]
-                subprocess.run(cmd, check=True, stderr=subprocess.PIPE, text=True, env=host_env())
-                logger.debug(f"Keyboard shortcut '{shortcut}' injected successfully")
-                return True
-            except subprocess.CalledProcessError as e:
-                logger.error(f"ydotool shortcut error: {e.stderr}")
-                return False
+            if not self._ensure_ydotoold():
+                logger.warning("ydotoold not ready before shortcut injection")
+
+            if self._ydotool_uses_legacy_named_keys():
+                # 0.1.x: one named chord token per step, all in a single call --
+                # it accepts any number of key sequences as positional args.
+                tokens = []
+                for modifiers, key in steps:
+                    token = self._ydotool_legacy_token(modifiers, key)
+                    if token is None:
+                        logger.error(
+                            f"No ydotool 0.1.x key name for '{key}' in shortcut '{shortcut}'"
+                        )
+                        return False
+                    tokens.append(token)
+                cmd = ["ydotool", "key"] + tokens
+            else:
+                # 1.x takes RAW KEYCODES only ("29:1 44:1 44:0 29:0"), never
+                # "ctrl+z" -- it documents non-interpretable values as "only cause
+                # a delay", so the old string form was a silent no-op.
+                seq = []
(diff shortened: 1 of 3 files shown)


src/vocalinux/text_injection/text_injector.py
```diff
@@ -1661,23 +1661,132 @@ def _inject_shortcut_with_wayland_tool(self, shortcut: str) -> bool:
...
+            for modifiers, key in steps:
+                codes = []
+                for name in modifiers + [key]:
+                    code = self._KEYCODES.get(name.lower())
+                    if code is None:
+                        logger.error(f"No keycode for '{name}' in shortcut '{shortcut}'")
+                        return False
+                    codes.append(code)
+                seq += [f"{c}:1" for c in codes] + [f"{c}:0" for c in reversed(codes)]
+            cmd = ["ydotool", "key"] + seq
```
jatinkrmalik: @eiseleb47 This always builds the 1.x raw-code form. After #734, main documents that the same form types digit garbage on 0.1.x, and ships `_ydotool_uses_legacy_named_keys()` plus the `_YDOTOOL_V1_*` / `_YDOTOOL_LEGACY_*` pairs for clipboard paste.

Please branch on that helper here too. Without it, voice undo/cut/copy/paste injects digits on common distro packages.

src/vocalinux/text_injection/text_injector.py
```diff
@@ -1661,23 +1661,132 @@ def _inject_shortcut_with_wayland_tool(self, shortcut: str) -> bool:
...
         else:
             logger.warning(f"Keyboard shortcuts not supported with {self.wayland_tool}")
             return False
 
+        try:
+            subprocess.run(cmd, check=True, stderr=subprocess.PIPE, text=True)
```
jatinkrmalik: @eiseleb47 Call `_wait_for_modifiers_released()` before this, and `_ensure_ydotoold()` when using the uinput helper.

Held PTT/toggle modifiers rewrite the chord. Flatpak 1.x needs the daemon; paste already ensures it.

jatinkrmalik (changes requested):
@eiseleb47 Thanks for raising this. The compositor-helper path and the `Home+shift+End` two-step parse look right, and failing loudly on an unmappable key is what we want.

I cannot merge it after #734. Main already has a dialect probe because raw 1.x codes type digits on 0.1.x packages. This helper always emits 1.x codes, skips the held-modifier wait, and never starts the daemon. Please fix those three, then we can land this before #716 and #714.

src/vocalinux/text_injection/text_injector.py
eiseleb47: Done — the ydotool branch now goes through `_ydotool_uses_legacy_named_keys()`.

One deviation from what you suggested: I didn't add `_YDOTOOL_V1_*` / `_YDOTOOL_LEGACY_*` constant pairs. Those work for paste because it's a single fixed chord; shortcuts vary per action, so 9 actions x 2 dialects would just re-encode `_KEYCODES`. Both dialects are built from `_parse_shortcut`'s steps instead, with `_ydotool_legacy_token()` for the 0.1.x form. Sanity check against your existing constants: it produces exactly `ctrl+v` and `ctrl+shift+v`.

The legacy names come from a separate `_YDOTOOL_LEGACY_NAMES` table rather than reusing `_KEYCODES`, and that turned out to be necessary — see the main comment. Short version: 0.1.x silently resolves an unknown name to its first character at exit 0, and `_KEYCODES` contains four names (`altgr`, `escape`, `space`, `return`) that 0.1.x doesn't have.

src/vocalinux/text_injection/text_injector.py
eiseleb47: Both added.

`_wait_for_modifiers_released()` sits before the tool branch rather than inside the ydotool one, so a held PTT modifier can't rewrite a wtype chord either — same placement as `_inject_with_wayland_tool`.

`_ensure_ydotoold()` warns and continues rather than aborting, matching what the paste path already does, since host 0.1.x often has no daemon at all.

Also put `env=host_env()` and a `timeout=` on the `subprocess.run` while I was here — the `host_env()` part was what the rebase conflicted on.

PR #716, opened 2026-08-20
fix(ibus): route keyboard shortcuts to a virtual-keyboard tool

eiseleb47:
## Description

`_inject_keyboard_shortcut` dispatched X11 and WAYLAND_XDOTOOL to xdotool and sent everything else to `_inject_shortcut_with_wayland_tool`, putting both IBus environments on a path that cannot serve them. IBus commits text; it has no way to synthesise a key combination.

WAYLAND_IBUS failed worse than it looked. `wayland_tool` is only set when dependency detection picks a Wayland injection tool, so on an IBus session it can be unset, and the attribute access raised inside the enclosing `try`. The `except` logs and returns `False`, so every shortcut reported a clean failure rather than a missing tool.

This routes X11_IBUS through xdotool, which reaches the same X server the XIM path is already using, and gives WAYLAND_IBUS a virtual-keyboard tool, preferring wtype and falling back to ydotool. With neither present the error names the real cause instead of surfacing an `AttributeError`.

X11_IBUS folds into the existing xdotool condition rather than adding a branch, since the call is identical. The `wayland_tool` write is taken under `_state_lock`, matching every other write to it; the injection call stays outside the block, as the lock is not reentrant.

Text injection is unchanged; this only affects shortcut delivery, and only in the two IBus environments.

## Type of Change

- [ ] 📖 Documentation update
- [ ] 🧹 Code refactoring
- [ ] ✅ Test update

## Checklist

- [ ] I have updated the documentation accordingly — no user-facing docs describe this behaviour
- [x] Pre-commit hooks pass

## Additional Notes

Five tests cover the routing: X11_IBUS reaching xdotool, WAYLAND_IBUS preferring wtype, its ydotool fallback, the no-tool-available failure injecting nothing, and an already-chosen tool being left alone.

---

### DIFF
--- a/src/vocalinux/text_injection/text_injector.py
+++ b/src/vocalinux/text_injection/text_injector.py
@@ -483,6 +483,39 @@ class TextInjector:
         )
         return "auto"
 
+    def _resolve_wayland_key_tool(self) -> Optional[str]:
+        """Pick (or reuse) the Wayland virtual-keyboard tool for key events.
+
+        Mirrors the preference ``_check_dependencies`` already applies: prefer
+        ydotool via uinput once its daemon is (or can be made) ready, and only
+        fall back to wtype. Callers that resolve a tool after startup -- IBus
+        shortcut routing, BackSpace key events -- use this instead of picking
+        their own order.
+
+        Returns:
+            The tool name, or None if neither is installed.
+        """
+        if getattr(self, "wayland_tool", None):
+            return self.wayland_tool
+
+        ydotool_available = shutil.which("ydotool") is not None
+        wtype_available = shutil.which("wtype") is not None
+
+        # _ensure_ydotoold can block for ~2s, so it stays outside the lock.
+        if ydotool_available and self._ensure_ydotoold():
+            tool = "ydotool"
+        elif ydotool_available and not wtype_available:
+            tool = "ydotool"
+        elif wtype_available:
+            tool = "wtype"
+        else:
+            return None
+
+        with self._state_lock:
+            self.wayland_tool = tool
+        logger.info(f"Using {tool} for Wayland key events")
+        return tool
+
     def _check_dependencies(self):
         """Check for the required tools for text injection."""
         ibus_requested = False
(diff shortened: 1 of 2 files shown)


src/vocalinux/text_injection/text_injector.py
```diff
@@ -1615,10 +1615,28 @@ def _inject_keyboard_shortcut(self, shortcut: str) -> bool:
...
-                return self._inject_shortcut_with_wayland_tool(shortcut)
+            if self.environment == DesktopEnvironment.WAYLAND_IBUS:
+                # Same on Wayland: fall through to a virtual-keyboard tool. Without
+                # this, self.wayland_tool may be unset -> AttributeError -> the
+                # except below swallows it and every shortcut silently returns False.
+                if not getattr(self, "wayland_tool", None):
+                    if shutil.which("wtype"):
+                        tool = "wtype"
+                    elif shutil.which("ydotool"):
+                        tool = "ydotool"
```
jatinkrmalik: @eiseleb47 Same compositor-helper-first rediscovery as #714. `_check_dependencies` prefers the uinput helper when the daemon is ready.

Please match that order. Also this still calls `_inject_shortcut_with_wayland_tool` below, which is broken until #715 lands with the dialect fix. Rebase onto that, and please do not stub the helper in the tests so CI cannot pass while IBus undo still fails.

jatinkrmalik (changes requested):
@eiseleb47 Thanks for raising this. Routing X11_IBUS through the X server path and giving WAYLAND_IBUS a real virtual-keyboard tool is the right diagnosis.

It still cannot land ahead of #715. This ends in `_inject_shortcut_with_wayland_tool`, which is still broken on main, and the new tests stub that helper so CI stays green while IBus shortcuts fail. Please rebase onto a fixed #715, match the existing uinput-first tool order, and keep a test that actually calls the real Wayland helper.

src/vocalinux/text_injection/text_injector.py
eiseleb47: All three addressed.

**Order** — the inline wtype-first block is replaced by `_resolve_wayland_key_tool()`, which mirrors `_check_dependencies`: ydotool once the daemon is ready, ydotool anyway if wtype is absent, then wtype. #714 now calls the same helper, so neither PR rediscovers the order.

**Rebase** — this sits on the fixed #715 rather than `main`, so the branch carries both commits until #715 merges. Needed for the point below.

**Stubbing** — the two tests that claim IBus shortcuts reach a tool now run the real `_inject_shortcut_with_wayland_tool` and assert its argv; only `shutil.which` and `subprocess.run` are patched. The three that assert *which* backend gets picked stay stubbed, since that's a different question, and `_stub_backends` says so in a docstring now.

Heads-up that `test_wayland_ibus_prefers_wtype` is renamed to `..._prefers_ydotool` — under the corrected order it prefers ydotool, so the old name described the bug rather than the fix.

PR #736, opened 2026-08-24
fix(installer): survive a release without a checksum manifest, and stop inventing GPUs

sesav:
## Description

`install.sh` is served from `main`, but it installs the latest **release tag**. Since #713 that means the script carries model verification while the tree it clones (v0.16.0) has no `model_checksums.txt`, so both pinned downloads failed closed, and VOSK fetched 40MB only to throw it away. Every user of the documented one-liner hits this today.

Fail-closed from #713 is kept. What changes is the one case it could not tell apart: a manifest that is **entirely absent** means the release predates pinning, and that release's app carries no runtime verification either, since `model_checksums.py` ships beside the manifest. Refusing to pre-download there keeps not one unverified byte off the disk; the app fetches the same files on first run, unchecked. It only costs the user their models. A manifest that is present but does not pin a model is untouched: still refused.

Same command, same machine (Debian 12, remote mode, tag v0.16.0):

| | before | after |
|---|---|---|
| `[ERROR]` in the log | 3 | **0** |
| VOSK download | 40MB, discarded | **none** |
| models | not installed | left to first run, explained once |

While reading those logs, four more defects turned up. The first is the most serious.

**A working GPU was never offered.** `install.sh` runs under `set -Eeuo pipefail` with `trap '… print_error …' ERR`, and `print_error` wrote to **stdout**. `vulkaninfo --features` exits non-zero on perfectly good drivers, and `FEATURES_OUTPUT=$(vulkaninfo --features …)` was unguarded — so the trap fired *inside* `$(check_vulkan_gpu_compatibility)` and its own message was captured as that function's return value. `VULKAN_COMPATIBLE` then never equalled `compatible`, `CAN_BUILD_GPU` came back `false`, and Step 3 told a machine with a supported GPU that *"GPU libraries not detected"*. The error never reached the log either — the substitution had swallowed it — which is why this looked like a phantom across three test installs.

Reproduced on Arch with Intel Lunar Lake, where `BACKEND_INFO` came back as:

```
cpu:No compatible GPU detected:false:yes:no:true:false:[ERROR] Unexpected error near line 124 …
compatible:[ERROR] Unexpected error near line 124 …
Intel(R) Graphics (LNL)
```

Diagnostics now go to stderr — `exec > >(tee …) 2>&1` keeps them in the log and on screen, but `$( )` no longer captures them — and the two assignments that treat "not compatible" as an answer rather than an error are guarded. Same mechanism explains the identical symptom on a Debian VM, where the trigger was the compat function's own `return 1`. This is pre-existing: `main` has the unguarded assignments, the stdout `print_error` and the ERR trap.

The same class bit `CUDA_TOOLKIT_ROOT=$(find_valid_cuda_toolkit_root)`, which could capture a warning into a path; moving diagnostics to stderr fixes that too.

**`detect_vulkan` treated any non-empty `vulkaninfo --summary` as a GPU.** The loader prints its header even when no ICD resolves, so a guest with no GPU announced one — named after the not-found fallback string, printed as though it were a device. It also read only `head -20` while `check_vulkan_gpu_compatibility` parsed the whole output, so on a machine whose device list starts later the two disagreed about the same hardware.

**Software renderers counted as GPUs.** The pattern list lived only inside `check_vulkan_gpu_compatibility`; `detect_vulkan` knew nothing about it. On a VirtualBox guest Step 1 promised *"Great performance with whisper.cpp Vulkan"* for `llvmpipe`, a CPU rasterizer on which Vulkan is slower than the CPU backend. One list now, read by both; Step 1 says `Software Vulkan only (llvmpipe …) - whisper.cpp CPU mode`.

**`HARDWARE_GPU_COUNT` dropped** — never read, and `((x++))` returns status 1 at zero, which under `set -e` would abort the function on the first hardware GPU.

**Build warnings: 15 → 0.** `project.license` as a TOML table and the License classifier are deprecated (setuptools removes them 2027-02-18) → SPDX `AGPL-3.0-only` + `license-files`, floor raised to `setuptools>=77`. Three resource directories added to `packages` to clear "Package would be ignored". Wheel contents are identical before and after — all 81 entries, diffed.

### What this does not do

It does not remove the skew, it makes it degrade cleanly. Models are still not pre-installed on a remote install until a release ships the manifest. Because `install.sh` is fetched from `main`, the fix reaches users without waiting for one.

## Related Issue

Part of epic #701 (Phase 2). Follow-up to #713 and #732.

## Type of Change

- [ ] 📖 Documentation update
- [x] 🧹 Code refactoring
- [x] ✅ Test update

## Checklist

- [ ] I have updated the documentation accordingly — no user-facing docs change
- [ ] Pre-commit hooks pass — see note below

## Additional Notes

**Verified on a Debian 12 VM and on an Arch laptop with a real Intel GPU:**

1. remote mode against v0.16.0 (the scenario above) — 0 errors, nothing downloaded, one clear message;
2. local repo mode on this branch — `ggml-tiny.bin verified (sha256)`, and the VOSK tree was accepted from its existing stamp, so 0 bytes fetched;
3. GPU detection — a diagnostic dumping both detectors' raw fields, on a Debian VM (software Vulkan) and on Arch with a real Intel GPU;
4. `just build` — 0 warnings, wheel unchanged.

**28 new tests** (22 + 6). Each was run against the unpatched `install.sh` first and fails there.

**Reviewer notes:**

- The **Flatpak** build is the only path consuming `pyproject.toml` without build isolation (`--no-build-isolation` against `org.gnome.Sdk` 50), so the `setuptools>=77` floor matters there. `flatpak.yml` triggers on `pyproject.toml` for `pull_request`, so this PR exercises it. AUR also builds `--no-isolation` but takes `python-setuptools` from Arch, well past 77.
- `AGPL-3.0-only` (not `-or-later`) matches how the project already describes itself everywhere; no "or later" wording exists in the repo.
- **Pre-commit** is unchecked because `isort` fails on `main` as well: `.pre-commit-config.yaml` runs `isort --profile black` while `justfile` lints with the same profile at line-length 100, and the two disagree on a pre-existing import in `tests/test_model_checksums.py`. Fixing that config mismatch is deliberately out of scope here.

**Known, deliberately not touched:** `pywhispercpp` is still installed twice — the unpinned range from `pyproject.toml` first, then the pinned `1.5.0`, which also re-downloads numpy. That is Phase 2's uv / `--require-hashes` item, a structural change to how the installer resolves dependencies.


jatinkrmalik (approved):
Thanks @sesav. Missing checksum file now skips the pre-download instead of fail-closing, and stderr plus the software-Vulkan filter stops a real GPU from being reported as missing. Merging.

PR #759, opened 2026-08-30
chore(release): make a release verifiable, and pin what builds it (epic #701, phase 5)

sesav:
## Description

v0.16.1 published four artifacts, two AppImages of ~100 MB, a wheel and an sdist with no checksum, no signature and no provenance. A user following our own instructions to download and run a 109 MB binary has nothing to check it against, in a project that pins the digest of all 67 models it downloads.

It could not have had one, either. The same version was built three times on three runners `build-and-release`, `build-appimage-arm64` and `publish-pypi` each ran their own `python -m build`, so the wheel on PyPI was never the wheel on the GitHub release, and a checksum for one would not have described the other.

**Epic:** #701, phase 5. Also closes phase 3 item 3.6, which shares this fix. Does not close the epic.

### What changed

**Built once, published everywhere.** `build-and-release` holds the only `python -m build` in the workflow, with `SOURCE_DATE_EPOCH` pinned to the tagged commit so a re-run reproduces the bytes. The aarch64 AppImage job and `publish-pypi` download that artifact instead of rebuilding — the wheel is `py3-none-any`, so the arm64 runner never needed one of its own, and no longer sets up Python at all: it needs a wheel and Docker.

**One manifest over everything.** A new `publish-checksums` job collects every artifact the run produced, writes `SHA256SUMS` and attaches it. It waits on both build jobs deliberately: the aarch64 AppImage lands after the release already exists, and a manifest covering three files out of four is worse than none, because the missing one is indistinguishable from a tampered one.

**Provenance from that same manifest.** `actions/attest-build-provenance` runs off `subject-checksums: dist/SHA256SUMS` rather than a second glob, so the file users check and the set we attest cannot drift apart.

**PyPI without a stored token.** `publish-pypi` uses trusted publishing; `PYPI_API_TOKEN` is gone from the workflow. This needs one-time setup on PyPI before the next tag — see below.

**Permissions per job.** The top level drops to `contents: read`. `id-token: write` was granted workflow-wide, which handed an OIDC token to `publish-aur` as well — the job that runs a third-party action with our AUR SSH key. Only `publish-pypi` and `publish-checksums` get it now. `packages: write` is gone; nothing used it.

**53 actions pinned by commit SHA** across all 8 workflows, in `@<sha> # v5` form. A tag is writable by whoever owns the action, so `KSXGitHub/github-actions-deploy-aur@v4.2.0` — the one handed the key that can push to our AUR package — was whatever that tag pointed at on the morning of a release. The single exception is `VocaHQ/.github/…@main`, our own org's shared workflow: the reason is a comment in the file, and the test allows a floating ref only for an owner we control, so the exemption cannot quietly grow to cover a third party.

`.github/dependabot.yml` is new, covering `github-actions` only, weekly, grouped into one PR. Pinning without an updater is how a pinned action stops receiving its own security fixes while the pin hides that it has gone stale.

The release notes now tell users what to do with all of this — `sha256sum -c` and `gh attestation verify` — and `docs/RELEASE_PROCESS.md` matches.

### Verified

Fourteen new tests, each checked by breaking the thing it guards:

```
extra `python -m build` in publish-pypi   → built-exactly-once, consumers-download-instead
publish-checksums stops waiting on arm64  → manifest-waits-for-every-artifact
id-token back at the workflow level       → only-the-jobs-that-need-it
SOURCE_DATE_EPOCH removed                 → build-timestamp-is-pinned
AppImage dropped from the sha256sum line  → manifest-covers-every-kind
attestation pointed at a glob instead     → provenance-from-the-manifest
back to twine + PYPI_API_TOKEN            → published-without-a-stored-token
publish-pypi loses id-token               → published-without-a-stored-token, and the above
`sha256sum -c` cut from the release notes → notes-tell-users-how-to-verify
one action returned to a tag              → third-party-action-is-pinned
a pin loses its `# v6` comment            → pin-records-the-version
a third-party action added on `@main`     → only-our-own-org-may-float
dependabot.yml deleted                    → something-moves-the-pins
```

All 8 workflows still parse. `pytest`: 2569 passed, 0 failed. black, isort and flake8 clean.

### Before this can ship

`publish-pypi` fails with `invalid-publisher` until PyPI knows about this workflow. On pypi.org (vocalinux/settings/publishing) add a GitHub publisher — owner `VocaHQ`, repository `vocalinux`, workflow `release.yml`, environment `pypi`. The `PYPI_API_TOKEN` secret can be deleted once the first trusted publish succeeds, not before. Written up in `docs/RELEASE_PROCESS.md` too.

### Not covered

`SOURCE_DATE_EPOCH` reaches the wheel and the sdist, not the AppImage: `docker-build.sh` forwards a fixed list of `-e` variables into the container and this is not one of them. The published checksum is correct — it is a re-run that would produce a different AppImage.

`nightly.yml` also publishes AppImages to a GitHub release with no checksums. Phase 5 scopes this item to `release.yml`, so nightly is left alone here rather than silently widened.

Pre-commit's `trailing-whitespace` hook fails on `nightly.yml` at four lines inside the heredoc that generates nightly release notes. Those four lines are identical on `main`; this branch neither introduced nor fixed them, because stripping whitespace inside that heredoc changes generated text.

Still open in phase 5: `just lock-check` only runs `uv lock --check`, so pinned versions in `requirements/*.txt` can still drift from `uv.lock` unnoticed. That is a separate PR — nothing calls `lock-check` today, so it needs a CI job of its own, and the six exports split across three different verification regimes (`uv export` is deterministic and currently in sync; `appimage*.txt` re-resolve but from exact pins; `whisper.txt` comes from `openai-whisper>=20231117` and cannot be checked by regeneration at all).

## Related Issue

Part of #701

## Type of Change

- [x] 🧹 Code refactoring
- [x] ✅ Test update

## Checklist

- [ ] Pre-commit hooks pass — passes on everything this branch touches; see the pre-existing `nightly.yml` note above

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,14 @@
+version: 2
+updates:
+  # Actions are pinned by commit SHA (tests/test_workflow_action_pins.py), which
+  # is only safe if something moves the pins: an action nobody bumps stops
+  # receiving its own security fixes, and the pin hides that it has gone stale.
+  # Grouped so this arrives as one pull request a week rather than fifteen.
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly
+    groups:
+      github-actions:
+        patterns:
+          - '*'
--- a/.github/workflows/codeberg-mirror.yml
+++ b/.github/workflows/codeberg-mirror.yml
@@ -17,7 +17,7 @@ jobs:
     runs-on: ubuntu-latest
     if: github.repository == 'VocaHQ/vocalinux'
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@fbc6f3992d24b796d5a048ff273f7fcc4a7b6c09 # v5
         with:
           fetch-depth: 0
 
--- a/.github/workflows/distro-test-matrix.yml
+++ b/.github/workflows/distro-test-matrix.yml
@@ -37,7 +37,7 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v5
+        uses: actions/checkout@fbc6f3992d24b796d5a048ff273f7fcc4a7b6c09 # v5
 
       - name: Install basic dependencies
         run: |
--- a/.github/workflows/flatpak.yml
+++ b/.github/workflows/flatpak.yml
@@ -51,10 +51,10 @@ jobs:
       options: --privileged
     steps:
       - name: Checkout
-        uses: actions/checkout@v5
+        uses: actions/checkout@fbc6f3992d24b796d5a048ff273f7fcc4a7b6c09 # v5
 
       - name: Build and bundle
-        uses: flatpak/flatpak-github-actions/flatpak-builder@v6
+        uses: flatpak/flatpak-github-actions/flatpak-builder@79327416609af08178ad73b352877e51450790b3 # v6
         with:
           bundle: vocalinux-${{ matrix.arch }}.flatpak
           manifest-path: packaging/flatpak/com.vocalinux.Vocalinux.yml
--- a/.github/workflows/issue-welcome.yml
+++ b/.github/workflows/issue-welcome.yml
@@ -13,4 +13,8 @@ permissions:
 
 jobs:
   welcome:
+    # Deliberately unpinned: this is our own org's shared workflow, inside the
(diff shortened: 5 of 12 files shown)


jatinkrmalik (approved):
@sesav LGTM!

PR #810, opened 2026-09-09
feat(ci): run install.sh in the distro matrix instead of grepping it

sesav:
## Description

The distro matrix never ran the installer. It grepped `install.sh` for function names, re-implemented `detect_distro()` inside the workflow, ran `bash -n` and `--help`, and created a venv by hand, all under `continue-on-error`.

**`scripts/install-test.sh`** (new) runs `install.sh --auto --skip-models` to completion in a distro container and checks what came out:

- as an unprivileged user with passwordless sudo, because `install.sh:677` refuses to run as root and calls `sudo` directly. That is why the job uses `docker run` from the runner rather than `container:`, which runs every step as root.
- from `git archive HEAD`, so the gate answers "does this commit install", and the mount stays read-only while local mode writes its venv into the container's copy.
- not `--skip-system-deps`: the per-distro package installation is the point. `--skip-models` because a model is 40-75 MB per container per run and that download is checksum-verified elsewhere.
- before the run it checks the tree still ships `src/vocalinux/utils/model_checksums.txt`, which `--skip-models` means nothing else in the gate reads.
- smoke: the venv interpreter, the `~/.local/bin/vocalinux` wrapper and its `--version`, the desktop entry, an import walk over main, tray, recognition and injection plus the Gtk and IBus typelibs, the `xclip` / `xsel` / `wl-copy` binaries, `ibus-daemon` on the distros whose package list promises it (arch, suse), and `EVDEV_AVAILABLE or PYNPUT_AVAILABLE`, since `keyboard_backends/__init__.py:34,40` imports both backends under `except ImportError`.
- on failure it dumps the last 200 lines of `~/.local/state/vocalinux/install-*.log`, the only place per-distro package manager output survives.

**`.github/workflows/distro-test-matrix.yml`**: job `test` becomes `install` and runs the gate; `continue-on-error` removed; `needs: syntax`; `timeout-minutes: 45`; `concurrency` with `cancel-in-progress`; paths filter gains `install.sh`, `uv.lock`, `requirements/**`, `vocalinux.desktop`. 273 lines of grep-based pseudo-tests removed.

**`justfile`**: `just install-gate <distro>` runs the same script locally (defaults to `debian:12`), same worktree handling as `aur-gate`. `AGENTS.md` lists it.

**`install.sh`, openSUSE**, both defects found by the gate's first run:

- `ibus-devel` requires `gtk-doc`, which zypper cannot satisfy once awk resolves to `busybox-gawk`: it offered an interactive solver choice, `--non-interactive` answered cancel, and the install died with `EXIT_MISSING_DEPS`. Replaced with `ibus typelib-1_0-IBus-1_0`. The headers were never needed, but the runtime package is: `ibus_engine.py:322` spawns `ibus-daemon -x -d -r` and `:350` shells out to `ibus engine`, and `ibus-devel` used to pull the daemon in transitively (`Requires: ibus = 1.5.34`).
- openSUSE had no compiler in its package list, where Ubuntu and Debian get `build-essential` and Arch gets `base-devel`. `python3-devel` supplies headers, not a compiler, and neither `evdev` nor `pyaudio` ships a wheel for the interpreter Tumbleweed carries. Added `gcc gcc-c++ make`.

**Tests**: `tests/test_installer_gate.py` (12 guards) holds the shape of the gate: the matrix runs the installer, the job can fail a build, `--skip-system-deps` is never passed, the tree comes from git rather than a copy, every matrix container has a bootstrap arm, both paths filters carry the installer, and the smoke asserts what a Python import cannot see. Guards match constructs (`[ -f "$TREE/src/vocalinux/utils/model_checksums.txt" ]`, `assert EVDEV_AVAILABLE or PYNPUT_AVAILABLE`) rather than bare substrings, because each of those paths and flags also appears in a comment or a `fail` message in the same file. `tests/test_cross_distro_compatibility.py` gains the `ibus` / `typelib-1_0-IBus-1_0` assertions on the `ZYPPER_PACKAGES` line it already inspects.

## Related issue

Part of #701 (phase 2.4). Does not close it: 2.1, the split of the 4803-line file, is the open half.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation
- [x] Tests
- [x] Packaging / CI

## Checklist

- [x] Code follows project style (Black, isort, flake8) - clean on both test files; `bash -n` clean on both scripts
- [x] Documentation updated when user-facing behavior or install steps change - `AGENTS.md` gains the gate; install steps are unchanged
- [x] Tests added or updated for the change - `tests/test_installer_gate.py`, plus the openSUSE assertions in `tests/test_cross_distro_compatibility.py`
- [x] Local tests pass (`pytest`) - full suite green
- [x] Conventional commit message style

N/A

## Notes for reviewers

**Verified by running it.** All six matrix containers green in CI, and all six run locally via `just install-gate` on the same installer and gate script: ubuntu 24.04 (153s), ubuntu 26.04 (174s), debian 12 (145s), fedora 42 (164s), arch (71s), tumbleweed (93s). Tumbleweed was red until both `install.sh` fixes landed; its zypper transaction now installs `ibus-1.5.34`. Every container reports `evdev=True pynput=False`, so the backend assertion is `or`, not `and`.

**Guards verified by mutation.** 21 plausible regressions each fail at least one guard: `continue-on-error` back, a container with no bootstrap arm, the `arch` arm deleted, `--skip-system-deps` added, `git archive` swapped for `cp -a`, `install.sh` dropped from one of the two paths filters or left only in a comment, the `ibus-daemon` or backend assertions removed, `ibus` dropped from `ZYPPER_PACKAGES`. Three cosmetic rewrites of `install.sh` fail none: `[[ $EUID -eq 0 ]]`, different quoting in the local-mode `grep`, reworded comments.

**The openSUSE package change alters what real users get.** `typelib-1_0-IBus-1_0` alone pulls `libibus-1_0-5` and no daemon, so the IBus injection path would silently stop being available on openSUSE; `ibus` requires `typelib(IBus)`, so the typelib comes with it either way and is listed explicitly.

**Not covered, deliberately:**

- Remote mode. The gate runs local mode only; `curl | bash` clones the latest tag rather than the commit under test, which is where #736 lived.
- Model downloads (`--skip-models`).
- The Wayland text-injection branch is uneven: session detection lands on `""` where the image has `loginctl` (arch, X11 arm only) and on `unknown` elsewhere (both arms), so which branch runs depends on the image.
- Fedora's `ibus-devel` pulls `ibus-libs` only and Ubuntu/Debian's `gir1.2-ibus-1.0` pulls `libibus` only, so neither installs the IBus daemon either. Left as is; the smoke records the split by requiring `ibus-daemon` only where the package list promises it.
- `resolve_install_tag` reports "Could not determine the latest release tag (GitHub API unreachable?)" when the real cause is that `curl` is missing. Harmless in local mode, misleading in general.

**Cost:** six containers in parallel, roughly 4 to 6 minutes each including the image pull, only on the paths filter.

**Merge gating:** the job can now go red, but the repository ruleset has no required status checks, so a red gate does not block a squash merge on its own.

### DIFF
--- a/.github/workflows/distro-test-matrix.yml
+++ b/.github/workflows/distro-test-matrix.yml
@@ -4,27 +4,38 @@ on:
   push:
     branches: [ main, master ]
     paths:
+      - 'install.sh'
       - '**/*.sh'
       - 'scripts/**'
       - 'pyproject.toml'
+      - 'uv.lock'
+      - 'requirements/**'
       - 'src/vocalinux/version.py'
+      - 'vocalinux.desktop'
       - '.github/workflows/distro-test-matrix.yml'
   pull_request:
     branches: [ main, master ]
     paths:
+      - 'install.sh'
       - '**/*.sh'
       - 'scripts/**'
       - 'pyproject.toml'
+      - 'uv.lock'
+      - 'requirements/**'
       - 'src/vocalinux/version.py'
+      - 'vocalinux.desktop'
       - '.github/workflows/distro-test-matrix.yml'
   workflow_dispatch:
 
+# Six containers a run, on a filter that any shell script matches: a second
+# push should not leave the first one's matrix running to completion.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
-  # The repo's only `bash -n install.sh` used to live inside the matrix below,
-  # under continue-on-error, so a syntax error in a 4698-line installer could
-  # not fail a build anywhere. This job is the one thing here allowed to fail.
-  # It is not phase 2.4 — nothing here runs the installer yet — it just stops
-  # the cheapest check we have from being advisory.
+  # First, because there is no point spending six containers on a script that
+  # does not parse.
   syntax:
     name: Shell syntax
     runs-on: ubuntu-latest
@@ -35,16 +46,21 @@ jobs:
           set -euo pipefail
           git ls-files -z '*.sh' | xargs -0 -n1 -t bash -n
 
-  test:
-    name: Test on ${{ matrix.container }}
+  install:
+    name: install.sh on ${{ matrix.container }}
     runs-on: ubuntu-latest
-    continue-on-error: true
+    needs: syntax
+    # A real install pulls the distro's whole GTK and build-toolchain closure
+    # before pip sees a single wheel.
+    timeout-minutes: 45
     strategy:
+      # Every distro reports independently: one failing arm is a finding about
+      # that distro, not a reason to lose the other five results.
       fail-fast: false
       matrix:
-        # README.md advertises Arch and Tumbleweed as tested and neither was
-        # here; fedora:39 has been EOL since 2024 while the AppImage boot
-        # matrix moved to 42. This list is what the docs promise.
+        # Adding a container here means adding a bootstrap arm to
+        # scripts/install-test.sh; tests/test_installer_gate.py holds the two
+        # lists together.
         container:
           - ubuntu:24.04
           - ubuntu:26.04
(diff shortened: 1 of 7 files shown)


jatinkrmalik (approved):
Yeah, this was on my list for a long time! @sesav thanks for picking it up! 🙏

Issue #699, opened 2026-08-17
[BUG] Keyboard layout still flips to US on GNOME/X11 — the `xkb:us::eng` guard in `get_current_engine()` is Wayland-only (follow-up to #664)


atistrcsn:
## Summary

On GNOME/X11 with a single non-US input source, Vocalinux permanently switches the X keymap to `us`. The system settings still show the correct layout (the GNOME indicat>

This is the same symptom as #664 (closed) and #292/#474/#497, but the root cause that survives on `main` is a different one: the `xkb:us::eng` sanity check in `get_curre>

## Environment

| | |
|---|---|
| OS | Ubuntu 24.04.4 LTS |
| Kernel | 7.0.0-28-generic |
| Session | **X11** (`XDG_SESSION_TYPE=x11`, `XDG_CURRENT_DESKTOP=ubuntu:GNOME`) |
| GNOME Shell | 46.0 |
| IBus | 1.5.29-rc2 (`ibus-daemon --panel disable --xim`) |
| Vocalinux | 0.15.0 (pip/venv install, autostarted via `~/.config/autostart/vocalinux.desktop`) |
| Layout | Hungarian (`hu`, QWERTZ) |
| Locale | `LANG=en_US.UTF-8`, `LC_TIME=hu_HU.UTF-8` |

## Actual behaviour

Three layers disagree after Vocalinux has been running:

```
$ localectl                 # system config
   X11 Layout: hu
   X11 Model: pc105

$ gsettings get org.gnome.desktop.input-sources sources
[('xkb', 'hu')]             # GNOME indicator says Hungarian

$ ibus engine
xkb:us::eng                 # <-- IBus is on US

$ setxkbmap -query
layout:     us              # <-- and so is the X server

$ xkbcomp -xkb $DISPLAY - | head -1
xkb_symbols "pc+us+inet(evdev)" {
```

Note there is only **one** XKB group and it is `us` — Hungarian is not loaded at all, so there is nothing to switch back to with `Super+Space`. Typing produces US charac>

## Expected behaviour

The layout stays `hu` — the layout GNOME and `localectl` both report.

## Root cause

`src/vocalinux/text_injection/ibus_engine.py` on `main` (8e308a1), `get_current_engine()`:

```python
# line 491
if engine == "xkb:us::eng" and _is_gnome_session() and _is_wayland_session():
```

[...]

PR #827, opened 2026-09-13
fix(ibus): stop the xkb:us::eng guard flipping GNOME/X11 to a US layout

AmirF194:
The gsettings fallback for the bogus `xkb:us::eng` default only checked
`_is_wayland_session()`. On GNOME/X11 the default is accepted as the real
engine and written back into the X keymap, which is what makes it
persistent there instead of just a display glitch. Dropped the Wayland
condition so the guard applies on GNOME regardless of session type.

Also fixed the second defect you found: `_get_gnome_current_source()` only
read `mru-sources`, so a session that never populated the MRU list (a
single input source, nothing switched yet) fell through to `None` even
with the guard fixed. It now falls back to the static `sources` list when
`mru-sources` comes back empty.

Added `test_get_current_engine_suspicious_us_default_on_gnome_x11` (fails
on main, passes on this branch) and
`test_empty_mru_sources_falls_back_to_configured_sources` for the second
defect. Full `tests/test_ibus_engine.py` suite: 128 passed, on 3.11 and
3.14. flake8/black/isort clean.

I don't have a real GNOME/X11 session to reproduce end to end, so this is
verified against the existing mocked-subprocess tests only, same as the
rest of this file's coverage.

### DIFF
--- a/src/vocalinux/text_injection/ibus_engine.py
+++ b/src/vocalinux/text_injection/ibus_engine.py
@@ -360,21 +360,18 @@ def _is_gnome_session() -> bool:
     return "GNOME" in desktop.upper()
 
 
-def _get_gnome_current_source() -> Optional[tuple[str, str]]:
-    """Return GNOME's current input source from the MRU list."""
-    if not _is_gnome_session():
-        return None
-
+def _read_gnome_input_sources_key(key: str) -> Optional[list]:
+    """Return the parsed list for a gsettings org.gnome.desktop.input-sources key, or None."""
     try:
         result = subprocess.run(
-            ["gsettings", "get", "org.gnome.desktop.input-sources", "mru-sources"],
+            ["gsettings", "get", "org.gnome.desktop.input-sources", key],
             capture_output=True,
             text=True,
             timeout=5,
             env=host_env(),
         )
         if result.returncode != 0:
-            logger.debug("gsettings get mru-sources failed; not using GNOME fallback")
+            logger.debug(f"gsettings get {key} failed; not using GNOME fallback")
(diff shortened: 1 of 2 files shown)


AmirF194:
This has been green for a week with no maintainer look yet. Scope is just ibus_engine.py: dropping the Wayland-only condition on the xkb guard, plus a fallback for _get_gnome_current_source when mru-sources is empty.

jatinkrmalik:
@AmirF194 thanks for the nudge. Sorry this sat. I went through tip `a348a8c` again for #699.

The two changes are the right ones:
1. Drop the Wayland-only gate so GNOME/X11 hits the same `xkb:us::eng` suspicion path as #497.
2. When `mru-sources` is empty, fall back to configured `sources[0]` so a fresh single-source session still recovers.

On the Greptile P1: I agree with your reply. Preferring `None` over trusting an unverified `xkb:us::eng` is the same tradeoff #497 already took on Wayland. Returning the live engine unchecked is exactly the bug this PR fixes. Not a merge blocker for me.

One tiny cleanup if you touch the branch again: a few debug/doc strings still say "GNOME/Wayland" after the condition covers all GNOME sessions. Pure log noise, not a behavior bug.

CI is green and the new X11 + empty-MRU tests look solid. I am ready to merge this. Want me to land it as-is, or do you want to push a one-line log cleanup first?

AmirF194:
Thanks for the careful re-read on the Greptile point, that's exactly the tradeoff. Good catch on the leftover "GNOME/Wayland" strings, I'll clean those up if I touch this file again.

PR #828, opened 2026-09-13
fix(deps): export and cap every engine extra

sesav:
## Description

`[parakeet]`, `[faster_whisper]` and `[vosk]` have no hash-pinned export and no upper bound, and `install.sh` installs all three by re-parsing `pyproject.toml` at install time, so they are the only engines a user can install with no hashes anywhere in the path.

- **`pyproject.toml`**: `sherpa-onnx<2`, `faster-whisper<2`, `vosk<0.4`. Re-locking moved no versions, only the three specifier lines in `uv.lock`.
- **`just lock`**: one `uv export` per engine extra, same flags as the vad export.
- **`requirements/{vosk,parakeet,faster-whisper}.txt`** (new): 27, 23 and 41 hash-pinned distributions, every one of them carrying at least one hash.
- **`scripts/check_exports.py`** (new), `just export-check`, and a step in the lint job: re-runs every `uv export` line in `just lock` and fails if a committed export is behind `uv.lock`. The lines are read out of the recipe, continuations included, so an export added there is checked without touching the script. `UV_OFFLINE=1`, so it needs neither the network nor the headers the test matrix installs. The `python` change filter gains `scripts/**`, so a checker-only change cannot skip the job that runs it.
- **`tests/test_dependency_exports.py`**: seven tests that enumerate rather than list. Every extra needs an export that exists, carries a hash on every pin, is regenerated by `just lock` and contains what the extra declares, and so does every extra `install.sh` installs, read out of its call sites, each of which has to yield a name. Two more require the drift check to cover every `uv export` line in the recipe, and the workflow's `python` filter to wake the job that runs it. `docs` is the one exemption and carries its reason in `EXPORT_EXEMPT`, checked against pyproject so a rename cannot hide behind it.
- **`tests/test_installer_review_notes.py`**: the expected vosk specifiers are read out of `pyproject.toml` rather than repeated as a literal.
- **`AGENTS.md`**: `One export per extra` and `Exports match the lock` rows.

## Related issue

Part of #701 (phase 1). Does not close it. Prerequisite for 2.3, which cannot install with `--require-hashes` from exports that do not exist.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation
- [x] Tests
- [x] Packaging / CI

## Checklist

- [x] Code follows project style (Black, isort, flake8)
- [x] Documentation updated when user-facing behavior or install steps change
- [x] Tests added or updated for the change
- [x] Local tests pass (`pytest`) - 3056 passed, 4 by-design skips
- [x] Conventional commit message style

N/A

## Notes for reviewers

**Verified by mutation.** Seventeen reintroduced defects each fail at least one test: an export deleted, or cut from `just lock`; hashes stripped from a whole export, from a single package, or from either of the two marker-split `numpy` blocks; a distribution cut from its own export; a new extra declared, the exempt one renamed, the exempt one left as the only extra, an extra declaring nothing; an undeclared extra installed under a quoted name, an installed name spelled with a dash, an installed name held in a variable; the drift check's step cut from the workflow, and `scripts/**` dropped from the change filter that wakes the job up, or moved into a neighbouring one. The checker itself fails on a hand-edited version, a deleted export and one removed `--hash=` line.

Reordering export lines fails nothing, by design. Removing a cap fails no test either, but not silently: `uv.lock` records the specifier, so dropping a bound without re-locking fails `uv lock --check` and the `--locked` flags the pipeline already uses.

The call sites are read strictly because `write_pip_reqs_skip_pygobject` matches `^<extra> = [` against `pyproject.toml` verbatim, writes an empty file when it misses, and `pip_install_reqs_file` returns 0 on an empty file: a name the test cannot read is an engine that installs nothing and reports success.

**Not covered, deliberately:** the exports are not consumed yet, `install.sh` still re-parses `pyproject.toml` and that swap is 2.3; neither engine reaches the AppImage or the Flatpak; `uv pip compile` targets stay out of the drift check, since `requirements/whisper.txt` and the AppImage files resolve against live indexes on purpose; `openai-whisper`, `torch`, `torchaudio` and `onnxruntime` stay uncapped, capping torch being a resolution decision rather than hygiene.

`just lock` reproduces the three files byte for byte on a cold `UV_CACHE_DIR` with no network, and the Flatpak manifest is untouched, since `sync_flatpak_deps.py` reads `runtime.txt` only.

**One local-only failure:** `tests/test_text_injector.py::test_wayland_wtype_fail_xdotool_when_uinput_blocked` fails on this branch and on clean `main` in the environment the run above came from. CI runs it on all four Python versions and it passes there. Not this change.


jatinkrmalik:
LGTM! Thanks @sesav

